PR #3, opened 2015-04-14
Bugfix: dep keys weren't invalidating CP upon change

machty:
Root cause: @computed('first', 'last') was actually
transpiling to a call to 
Ember.computed(['first', 'last'], fn)


rwjblue:
Thank you!!

PR #7, opened 2015-04-14
a-bide by arity. If you don’t ask for it, you won’t get it.

stefanpenner:
this is more controversial then the last PR, but would love to get some thoughts/eyes on it.

It is possible for a CP to cause an invalidation, but to never be recomputed if it is never `get`'d. By always  getting CPs even if no arg for them is used, we do somewhat change the semantics.

Unfortunately, and likely why this should be merged is restArgs and arguments hacks wont work.

```js

var a = {
  @computed('firstName', 'lastName')
  foo(...names) {
    return names.filter(Boolean).join(' ');
  }
};
```

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -18,8 +18,10 @@ function handleDescriptor(target, key, descriptor) {
 }
 
 function callUserSuppliedFunction(params, func) {
+  let arityAlignedParams = params.slice(0, func.length);
+
   return function() {
-    let paramValues = params.map(p => get(this, p));
+    let paramValues = arityAlignedParams.map(p => get(this, p));
 
     return func.apply(this, paramValues);
   };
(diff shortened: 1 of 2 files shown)


rwjblue:
Not sure about this one. I think I'd rather produce another export that does this so that you can choose which way you want.

Possibly even use two decorators in tandem:

```
var a = {
  @ignoredDependentKeys('middleName')
  @computed('firstName', 'lastName', 'middleName')
  foo(...names) {
    return names.filter(Boolean).join(' ');
  }
};
```

What do you think?

stefanpenner:
ya, im not sure. Mostly I realized I wanted this to ignore stuff like

```
 @computed('firstName', 'lastName', 'some.@each.fpp}massive,expansion}')
```

PR #11, opened 2015-04-16
add other computed macros

stefanpenner:
(this may actually work, pending spec people telling me if i am wrong or babel is wrong)

- [X] need to report babel issue (and get the spec to add further clarification)
- [x] pending babel fix https://github.com/babel/babel/issues/1277
- [x] once the babel fix lands, we will need to update this project to take into account descriptor initializers

```js
export default Ember.Object.extend({
  init() {
    this._super(...arguments);
    this.first = 'Rob';
    this.last = 'Jackson';
  },

  @alias('first') firstName,
  @empty('first') missingFirstName,
  @match('first', /rob/i) isRob
});
```

- [x] ember-data

```js
export default DS.Model.extend({
  @attr firstName,
  @attr({ defaultTo: 'smith' }) lastName,
  @hasMany cars,
  @belongsTo human
});
```

### DIFF
--- a/addon/utils/handle-descriptor.js
+++ b/addon/utils/handle-descriptor.js
@@ -1,0 +1,50 @@
+import Ember from 'ember';
+import computed from 'ember-new-computed';
+
+const { get } = Ember;
+
+export default function handleDescriptor(target, key, desc, params = []) {
+  return {
+    enumerable: desc.enumerable,
+    configurable: desc.configurable,
+    writeable: desc.writeable,
+    initializer: function() {
+      let computedDescriptor;
+
+      if (desc.writable) {
+        var val = desc.initializer();
+        if (typeof val === 'object') {
+          let value = { };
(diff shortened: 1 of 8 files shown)


stefanpenner:
* npm cache needs to be purged
* 1.10 is failing, but I believe it was failure before

addon/utils/handle-descriptor.js
```diff
@@ -0,0 +1,50 @@
...
+  return function() {
+    let paramValues = params.map(p => get(this, p));
+
+    return func.apply(this, paramValues);
+  };
+}
+
+
+function callUserSuppliedSet(params, func) {
+  return function(value) {
```
martndemus: this needs to be `(key, value)`

rwjblue:
:shipit:

rwjblue:
Released as v0.1.3

PR #26, opened 2015-07-10
Update to newest ember

martndemus:
Also trying to fix all broken tests, the readOnly test is a though one I haven't cracked yet. So this is still WIP.


martndemus:
The problem seems to be that `@readOnly` and `@computed` need to be ordered.
If the order is wrong, then the descriptor's value/initializer is not yet a CP and thus readOnly trips up, or writeable is false and thus handleDescriptor throws an error.

stefanpenner:
Ya, the issue is. Babel originally implemented this flow incorrectly (backwards i believe), and this lib was implemented to work in-light of that issue.

Basically, I  wrote knowingly wrong code, in-order to try this out. But haven't had time to get back to fix it after babel fixed itself.

Anyways, I would love help :)

stefanpenner:
can we split the PR into 2 pieces

1. upgrade ember (yes there will be failures)
2. fix this addon?

martndemus:
Separate pr for fixes is on its way.

Issue #20, opened 2015-05-20
should @computed('attr.something.@each.foo') provide attr.something?


stefanpenner:
example:

``` js
@computed('attr.model.@each.{something,somethingElse}')
foo(model) {
  // I think we want model === attr.model
  // not model == some useless proxy object
}
```

thoughts?

rwjblue:
`@computed('attr.model.@each.{something,somethingElse}')` should pass the value of `attr.model` in to the callback function.

We should just pass in the portion prior to the `.@each` or `.[]`.

stefanpenner:
> We should just pass in the portion prior to the .@each or .[].

my thoughts exactly +1

stevesims:
I ran into this behaviour just yesterday, and frankly I was rather surprised to discover that I wasn't getting the value passed through to my computed function that I had specified as my dependent key.  The behaviour exhibited was non-obvious to me and unexpected.

Yes, I do realise that the behaviour from this issue is as has been documented, but I had skipped past that part of the documentation before.

My presumption was that by specifying it as the key I'd get passed `something.@each.foo`, i.e. an array of those `foo` values.  The first statement of my CP function ended up being a `get` for that key, and I ignored the `something` being passed into my function.

Had I wanted to work with `something`, but make sure that the CP is affected by the `foo` values changing then intuitively I'd have written my declaration to be `@computed('something', 'something.@each.foo') bar(something) {`.

pzuraq:
I think this is now more of a concern with kellyselden/ember-macro-helpers, as this lib now just wraps that one as a decorator

PR #29, opened 2015-07-13
Pass in the part before @each/[] as attr for a CP

martndemus:
Solves #20.

I also have a commit ready for `foo.{bar,baz}`, that it will pass in `foo`, but I was unsure if that is the right thing to do. The alternative is to pass in `bar` and `baz`.

### DIFF
--- a/addon/utils/handle-descriptor.js
+++ b/addon/utils/handle-descriptor.js
@@ -35,7 +35,11 @@ function getAttr(obj, name) {
   let i;
 
   for (i = 0; i < parts.length; i++) {
-    if (parts[i] === '@each' || parts[i] === '[]') { break; }
+    if (parts[i] === '@each' ||
+        parts[i] === '[]'    ||
+        parts[i].indexOf('{') !== -1) {
+      break;
+    }
   }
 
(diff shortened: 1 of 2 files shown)


rwjblue:
> I also have a commit ready for foo.{bar,baz}, that it will pass in foo

:+1:

rwjblue:
Tests for beta and canary are failing because of ember-data, should be fixed by https://github.com/rwjblue/ember-computed-decorators/pull/30.

rwjblue:
Can you add support for `foo.{bar,baz}` passing `foo`, and rebase (to get the fixes from #30)?

martndemus:
@rwjblue done!

rwjblue:
Could we do the string processing outside of the returned function (instead of every time the end users `get`/`set` callbacks are called)?

The current implementation uses `getAttr` inside the function that is calling the users callback, if instead we update the `params` array just inside `callUserSuppliedGet` (but before the returned callback) we avoid the extra string manipulation and looping for every call to `get`.

Thoughts?

martndemus:
I agree! (and updated the code)

rwjblue:
Awesome, thank you!

PR #33, opened 2015-07-19
Update aliases for CP methods

martndemus:
Also sorted them alphabetically so I could better compare what missed with the API docs.


rwjblue:
:+1:

stefanpenner:
:

martndemus:
@stefanpenner see #34

PR #39, opened 2015-09-03
Add test for @computed with nested property

### LINKED ISSUE #37: Add test for @computed passing in nested property
Per discussion with @rwjblue and [another issue](https://github.com/rwjblue/ember-computed-decorators/issues/20#issuecomment-103964356), we should expect `@computed('foo.bar')` to pass the value of `foo.bar` to the callback function. There is currently no test for this (though [there are for `attr.models.@each` and `attr.models.[]`](https://github.com/rwjblue/ember-computed-decorators/blob/master/tests/unit/computed-test.js#L165-L213)).

SaladFork:
Closes #37

### DIFF
--- a/tests/unit/computed-test.js
+++ b/tests/unit/computed-test.js
@@ -162,6 +162,31 @@ test('works properly', function(assert) {
   assert.equal(callCount, 1);
 });
 
+test('attr.foo passes attr.foo', function(assert) {
+  assert.expect(2);
+
+  let obj = {
+    attr: {
+      foo: 'bar'
+    },
+
+    /* jshint ignore:start */
+    @computed('attr.foo')
+    /* jshint ignore:end */
+    something: {
+      get(foo) {
+        assert.deepEqual(foo, 'bar');
+      },
+      set(value, foo) {
+        assert.deepEqual(foo, 'bar');
+      }
+    },
+  };
+
+  get(obj, 'something');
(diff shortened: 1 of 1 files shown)


SaladFork:
I'm unsure how to address the Travis failure, =\
```
bower ECONFLICT     Unable to find suitable version for ember
The command "bower install" failed and exited with 1 during .
```

SaladFork:
Once this PR is merged I'll submit another PR that moves a bunch of the tests from the "decorated computed without dependent keys'" that have dependent keys to "decorated computed with dependent keys".

rwjblue:
You should be able to reproduce the issue locally with `ember try ember-release test`.  I think we need to add resolution values?

SaladFork:
```
Unable to find a suitable version for ember, please choose one:
    1) ember#>=1.4 <2 which resolved to 1.13.3 and is required by ember-cli-shims#0.0.3, ember-load-initializers#0.1.5
    2) ember#1.13.3 which resolved to 1.13.3 and is required by ember-computed-decorators
    3) ember#> 1.5.0-beta.3 which resolved to 1.13.3 and is required by ember-resolver#0.1.21
    4) ember#^2.0.0 which resolved to 2.0.1 and is required by ember-data#ee0ccd74e7
```

I suspect the project's `bower.json` will need to be updated.

rwjblue:
Submitted #40

Issue #25, opened 2015-07-03
Use machty's jshint guard hax


machty:
There might be some not-yet-handled edge cases but I think this should do the trick:

https://github.com/machty/broccoli-es7-decorator-jshint-guards

@rwjblue and I talked about this and he suggested using and re-exporting the above in ember-computed-decorators. I was gonna do this and then got busy (and I don't totally grok the test failures due to the recent Babel change).

stefanpenner:
https://github.com/machty/broccoli-es7-decorator-jshint-guards/blob/master/index.js#L11 doesn't look sourceMap aware.

rwjblue:
This transformation is done before any transpiration, which means that those lines will just be shown in the sourcemap output.

stefanpenner:
ah yes, but they will be different from the input. It would be better if they where done as a babel plugin, that way it would be totally transparent.

martndemus:
@stefanpenner Are there any babel plugins that do similar things? I don't think I can build something like this from scratch, but if there is something I can 'copy' from, then I might manage to do this.

devinus:
I feel like instead ember-cli should switch to ESLint and use a ESLint plugin to understand decorators.

stefanpenner:
> I feel like instead ember-cli should switch to babel-eslint to understand decorators.

this is likely the best solution for right now.

SirZach:
any chance on using @machty code as a temporary means? not sure how far away #49 is to being completed and i would love to rip out all the hax. thanks!

rwjblue:
#49 is only for updating this Addons linter to ESLint, and doesn't fix the problem for applications.

rwjblue:
Also, in general I'm not opposed, but we need to figure out how to address the sourcemap concerns that @stefanpenner mentioned first.

devinus:
I've updated my apps to use ESLint. It's glorious.

SirZach:
@devinus could you go into more detail? do you not have jshint anymore and instead have eslint?
@rwjblue sometimes you gotta do something bad and ask for forgiveness later :)

pzuraq:
Think this might be superceded by #106, which upgrades to ESLint

PR #49, opened 2015-11-06
[WIP] Switch from JSHint to ESLint

martndemus:
So we don't have to use the annoying JSHint guards everywhere
Addresses #25

To-Do:
- [x] Get ember-cli/ember-cli-qunit#87 merged
- [ ] Submit patches to linters that add rules specific to decorators (babel/eslint-plugin-babel#28)


martndemus:
After some toying with the supposed faulty `unused variable` warnings. I think the warnings are actually correct.

Take the following example:

```javascript
var foo = 42;
var obj = { foo };
obj.foo; 
// => 42
```

Which isn't really different from

```javascript
var foo = 42
var obj = { @readOnly foo }
obj.foo
// => 42
```

So things like:
```javascript
var Model = DS.Model.extend({
  @attr firstName,
  @attr({ defaultTo: 'blue' }) lastName
});
```

can cause subtle bugs.

So although it doesn't look that great, it would be more correct to write:
```javascript
var Model = DS.Model.extend({
  @attr firstName: null,
  @attr({ defaultTo: 'blue' }) lastName: null
});
```

SirZach:
bump?

martndemus:
Thanks for reminding me, I'll try to work on this soon™

pzuraq:
This was superceded by #106

PR #61, opened 2016-02-23
Add failing test cases for ES6 classes

offirgolan:


Ramblurr:
Are there any plans to support es6 classes soon?

topaxi:
I think ES6 class support is nearly there, but they don't make sense yet, as the decorator proposal is still in flux. Babel 6 even dropped decorator support: http://babeljs.io/docs/plugins/transform-decorators/.

Ramblurr:
Interesting. @topaxi do you see any future headaches by using this addon now?

topaxi:
@Ramblurr I definitely see future hazards if you are using this now, first, we don't know when decorators will be standardized, second, we don't know what changes will be made until then. As far as I know, in the current spec, the order in which the decorators are executed was reversed, which might be a problem sooner than later (I'm not 100% sure on this one though @rwjblue might know more here).

rwjblue:
@offirgolan - Mind rebasing here?  I believe these tests will pass...

tests/unit/computed-test.js
```diff
@@ -266,3 +266,47 @@ test('attr.@each.{foo,bar} passes attr', function(assert) {
...
+  class Foo {
+    constructor() {
+      this.first = 'rob';
+      this.last = 'jackson';
+    }
+
+    /* jshint ignore:start */
+    @computed('first', 'last')
+    /* jshint ignore:end */
+    get fullName() {
```
rwjblue: We explicitly throw on this:

> ember-decorators does not support using getters and setters

I think mostly because we haven't thought it through just yet.

Thoughts @stefanpenner  / @pzuraq?

pzuraq: I think we can do this, we just have to figure out how to reach into existing descriptors in cases where both a getter and setter are used. I'll try to find some time later today to spike it out.

rwjblue: Ya, I think it makes sense for `@computed` (for example), but not all of them.  e.g. `@observer` with a getter or setter is pretty WAT...

pzuraq: Ah, yeah definitely, this should probably only be for `@computed` (can't think of another one where it would make sense tbh)

PR #72, opened 2016-09-23
Updating documentation for addons to not throw deprecation

gmurphey:
Not calling `this._super` in the `init` hook for addons throws deprecations in consuming applications. Made a simple update to the addon snippet for babel decorator support in the readme.


rwjblue:
Thank you!

PR #76, opened 2016-11-15
ember-cli@2.9.1

kellyselden:
I couldn't get the tests to run locally. After this they now run and pass.


rwjblue:
Thanks @kellyselden!

PR #77, opened 2016-11-15
move out expandPropertyList to be shared by others (ember-awesome-macros)

kellyselden:
for those watching, after chatting with @rwjblue, we are going to try and combine some duplicate code from both here and https://github.com/kellyselden/ember-awesome-macros.

re:
https://github.com/kellyselden/ember-awesome-macros/pull/120
https://github.com/kellyselden/ember-macro-helpers/pull/1


rwjblue:
Thanks @kellyselden!

PR #100, opened 2017-06-18
refactor(macros): Refactors macros and adds inject decorators

pzuraq:
This PR does two things:

1. Implements decorators for `service` and `controller`
2. Refactors the various macros and wrappers that produce decorators to normalize them across the codebase and make it easier to add more and understand what they are doing

The two tasks could be broken out into separate PRs, and tests still need to be written for the new features and some new tests written for edge cases that weren't caught before (the `errorMessage` for `observes` and `on` would never fire, for instance). Mainly submitting this for feedback and to make sure this is a decent direction to take this in.


addon/utils/is-descriptor.js
```diff
@@ -1,7 +1,6 @@
 export default function isDescriptor(item) {
   return item &&
     typeof item === 'object' &&
-    'writable' in item &&
```
pzuraq: This one was weird, could be a bug with Babel, but in a consuming Babel 6 app, when I ran a decorator on a class property with no parameters, the descriptor didn't have the writable property at all:

```
class FooComponent extends Component {
  @service bar
}
```

addon/inject.js
```diff
@@ -0,0 +1,5 @@
+import Ember from 'ember';
+import { decoratorWithKeyReflection } from './utils/decorator-macros';
+
+export var service = decoratorWithKeyReflection(Ember.inject.service);
```
rwjblue: `const` 😸 
`var` :rage4:

rwjblue:
Thank you @pzuraq!

PR #105, opened 2017-06-22
docs(readme): Fix typographical error

Kerrick:
`Gannon` -> `Ganon`, because `Gannon` is only used in the first few Legend of Zelda games, but the presence of Navi implies we're talking in more modern terms.


rwjblue:
😹 🕹

PR #107, opened 2017-06-28
chore(docs): Adds ESDoc

pzuraq:
The README is getting a bit difficult to maintain so I thought I would take a stab at adding API docs to this library. Unfortunately, the Ember standard YUIDoc has no notion of ES6 modules or static functions, and since this library consists entirely of them it doesn't really fit.

Out of the other options out there, the top 3 were:

* JSDoc
* ESDoc
* TypeDoc

TypeDoc is ideal because it does 90% of the documentation for you and seems to be the agreed upon standard solution for Typescript, but would require switching to TS and figuring out some kind of build system (or forcing users to download the entire TS compiler). This may be worth it, but I didn't want to make the jump without getting feedback first.

We're currently using ESDoc at Addepar. It has a forward thinking mentality where, like TypeDoc, it pulls as much information from the code itself as possible, meaning the documentation itself is fairly minimal. However, the project is fairly young, and currently only has one maintainer/contributor who seems to be a little sporadic. Hopefully it will open up more as it matures, and it could be a great stand-in for TypeDoc on non-Typescript projects.

JSDoc seems to be moving faster than YUIDoc, but doesn't have support for experimental features and is fairly lacking in terms of documentation generation. If ESDoc is too experimental and TypeDoc/Typescript too difficult to pursue, it will work, but I think we could do better.

Screenshots of the rendered documentation:

screen shot 2017-06-27 at 10 15 30 pm

screen shot 2017-06-27 at 10 15 46 pm

screen shot 2017-06-27 at 10 16 08 pm


rwjblue:
Looks great, thank you for working on this! What do I need to do after merging to get the docs published?

pzuraq:
Just added [ember-cli-github-pages](https://github.com/poetic/ember-cli-github-pages) and tested it out on my fork, once merged you should be able to follow the steps on the README there and everything should Just Work ™️

PR #108, opened 2017-06-28
feature(action): Adds @action decorator

pzuraq:
Adds an `@action` decorator, which works by creating a passthrough function on the `actions` hash of the decorated class.

### DIFF
--- a/addon/action.js
+++ b/addon/action.js
@@ -1,0 +1,45 @@
+import { decorator } from './utils/decorator-wrappers';
+import extractValue from './utils/extract-value';
+
+/**
+ * Decorator that turns the target function into an Action
+ *
+ * Adds an `actions` object to the target object and creates a passthrough
+ * function that calls the original. This means the function still exists
+ * on the original object, and can be used directly.
+ *
+ * Before:
+ *
+ * ```js
+ * export default Ember.Component.extend({
+ *   actions: {
+ *     foo() {
+ *       // do something
+ *     }
+ *   }
+ * });
+ * ```
+ *
+ * After:
+ *
+ * ```js
+ * export default MyComponent extends Ember.Component {
+ *   @action
+ *   foo() {
+ *     // do something
+ *   }
+ * }
+ * ```
+ */
+export default decorator(function(target, key, desc) {
+  const value = extractValue(desc);
+
+  if (typeof value !== 'function') {
+    throw new Error('The @action decorator must be applied to functions');
+  }
+
+  target.actions = target.actions || {};
+  target.actions[key] = value;
+
+  return value;
+});
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -4,9 +4,11 @@ const EmberAddon = require('ember-cli/lib/broccoli/ember-addon');
 module.exports = function(defaults) {
   var app = new EmberAddon(defaults, {
     babel: {
-      plugins: ['transform-decorators-legacy']
+      plugins: [
+        'transform-decorators-legacy',
+        'transform-class-properties'
+      ]
     },
-
     esdoc: {
       enabledEnvironments: ['production'],
       config: {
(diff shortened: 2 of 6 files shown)


tests/unit/action-test.js
```diff
@@ -0,0 +1,50 @@
...
+    @action
+    foo(val) {
+      if (val === 'bar') {
+        assert.ok(true, 'action called through send');
+      } else if (val === 'baz') {
+        assert.ok(true, 'action called directly');
+      }
+    },
+
+    send(fn, ...args) {
```
pzuraq: Probably better to use legit action sending, just didn't have the time to figure out the best way to do it 😕

rwjblue: I'd say the easiest would be to make a "simple" component integration test instead.

```js
moduleForComponent('actions', { integration: true });

test('action decorator', function(assert) {
  this.register('component:foo-bar', class extends Ember.Component {
    @action
    foo() {
      assert.ok(true, 'called!');
    }
  });
  this.register('template:components/foo-bar', hbs`<button {{action 'foo'}}>Click Me!</button>`);

  this.render(hbs`{{foo-bar}}`);

  return click('button');
});
```

rwjblue: You could also include the mixin that provides it in Ember, but I don't think you can reliably include the mixin in `class` syntax...

addon/action.js
```diff
@@ -0,0 +1,47 @@
...
+ *     // do something
+ *   }
+ * }
+ * ```
+ */
+export default decorator(function(target, key, desc) {
+  const value = extractValue(desc);
+
+  if (typeof value !== 'function') {
+    throw new Error('The @action decorator must be applied to functions')
```
rwjblue: missing semi

rwjblue: (we should fix eslint to ensure it barks about this)

addon/action.js
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  target.actions = target.actions || {};
+  target.actions[key] = function() {
```
rwjblue: Thoughts on doing `target.actions[key] = value;` instead?  It would avoid the extra closure, and _seems_ like it would do the trick still....

tests/unit/action-test.js
pzuraq: Ah for sure, didn't realize you could dynamically register that easily. Will update them!

addon/action.js
pzuraq: 👍

addon/action.js
pzuraq: Yeah that makes sense, could also sidestep some weird `this` context behavior (in the tests it actually ended up referring to the `actions` object, which I know isn't normal but still)

pzuraq:
re: Optional name param - Do we have a compelling use case for that? I'm always hesitant to add to the public API when all it does is add more ways to do the same thing 😛 but it's not entirely out there, I could see people wanting it for one reason or the other.

ember-cli-build.js
```diff
@@ -4,7 +4,10 @@ const EmberAddon = require('ember-cli/lib/broccoli/ember-addon');
 module.exports = function(defaults) {
   var app = new EmberAddon(defaults, {
     babel: {
-      plugins: ['transform-decorators-legacy']
+      plugins: [
+        'transform-decorators-legacy',
+        'transform-class-properties'
```
pzuraq: Added this for the new test that checks that the error is thrown if `@action` is applied to a non-function. I'll be submitting another PR to update the README and include tests/examples for ES6 classes soon.

rwjblue:
> re: Optional name param - Do we have a compelling use case for that?

Nope! I agree, we should only consider adding it if/when we have a compelling use case...

addon/action.js
rwjblue: ya, Ember's current behavior is to grab the function `let action = obj.actions[actionName]` then use `.apply` to call it with the correct `this` binding.  With the way the tests were originally, I suspect you would have had to emulate that in the test environment

Issue #98, opened 2017-06-09
Ember Data decorators do not appear to work with ES6 classes


ebryn:
```
import DS from 'ember-data';
import { attr } from 'ember-computed-decorators/ember-data';

export default class extends DS.Model {
  @attr title;
};
```

Using Babel 6 with this EmberApp config:
```
    babel: {
      plugins: [
        'transform-decorators-legacy',
        "transform-class-properties"
      ]
    }
```

pzuraq:
I think this is blocked by some pending changes in Ember Data

cc @runspired

NullVoxPopuli:
decorators currently don't work at all with ember-cli-babel 6. 
see:
https://github.com/rwjblue/ember-computed-decorators/issues/96
https://github.com/rwjblue/ember-computed-decorators/issues/93
https://github.com/rwjblue/ember-computed-decorators/issues/94
https://github.com/loganfsmyth/babel-plugin-transform-decorators-legacy/issues/65
and kinda: https://github.com/babel/ember-cli-babel/issues/69

People seem to be content with adding `: null` to all their decorators. :-1: 

I've tried `babel-plugin-transform-decorators` -- also doesn't work.

I have yet to find a way to get decorators working without adding 130 `: null`'s

pzuraq:
The null fix is for the standard Ember object model, when you are using `extend`. When used in ES6 classes null isn't needed, and it works generally, but not for DS.Model yet.

NullVoxPopuli:
does this only work with components?
I tried:
```js
export default class CurrentUserService extends Ember.Service { ...
``
and I get a Broccoli error

**Comment by rwjblue** [comment]
Not sure why you got a syntax error, but it does work generally speaking...

**Comment by NullVoxPopuli** [comment]
@rwjblue do you have an opensource project that I can take a gander at dependencies and configs and see if I can find what I'm missing?

**Comment by NullVoxPopuli** [comment]
found out I needed babel-plugin-transform-class-properties. everything is working okish, and everything is more modern now :-)

**Comment by runspired** [comment]
I don't have to use : null to use decorators.. likely a config issue here

**Comment by pzuraq** [comment]
With babel 6 and the `transform-decorators-legacy` plugin, properties must have a value when using decorators on an object:

```js
export default Ember.Object.extend({
  @alias('foo') bar, // invalid
  @alias('foo') baz: null, // valid

  @computed
  foo() {
    // valid, because the function is the value
  }
});
```

[...]

PR #112, opened 2017-07-02
Attempt at failing test for es6 ember-data relations

NullVoxPopuli:
ref: https://github.com/rwjblue/ember-decorators/issues/98#issuecomment-312515592

### DIFF
--- a/tests/unit/ds-macro-test.js
+++ b/tests/unit/ds-macro-test.js
@@ -37,3 +37,32 @@ test('DS macro', function(assert) {
 
   assert.deepEqual(relationships, ['user', 'car']);
 });
+
+class ES6Model extends DS.Model {
+  @attr firstName;
+  @attr({ defaultTo: 'blue' }) lastName;
+  @hasMany user;
+  @belongsTo car;
+}
+
+test('DS macro (ES6)', function(assert) {
+  ES6Model.store = {
+    modelFor(typeKey) {
+      return typeKey;
+    }
+  };
+
+  var attributes = [];
+  ES6Model.eachAttribute(function(attr) {
+    attributes.push(attr);
+  });
+
+  assert.deepEqual(attributes, ['firstName', 'lastName']);
+
+  var relationships = [];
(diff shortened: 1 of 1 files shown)


tests/unit/ds-macro-test.js
```diff
@@ -37,3 +37,32 @@ test('DS macro', function(assert) {
 
   assert.deepEqual(relationships, ['user', 'car']);
 });
+
```
pzuraq: You should probably define the class inside the body of the test itself, see the `@action` decorator test for an example

tests/unit/ds-macro-test.js
```diff
@@ -37,3 +37,32 @@ test('DS macro', function(assert) {
...
+class ES6Model extends DS.Model {
+  @attr firstName;
+  @attr({ defaultTo: 'blue' }) lastName;
+  @hasMany user;
+  @belongsTo car;
+}
+
+test('DS macro (ES6)', function(assert) {
```
pzuraq: You should probably define the class inside the body of the test itself, see the `@action` decorator test for an example

tests/unit/ds-macro-test.js
```diff
@@ -37,3 +37,32 @@ test('DS macro', function(assert) {
 
   assert.deepEqual(relationships, ['user', 'car']);
 });
+
+class ES6Model extends DS.Model {
```
pzuraq: You should probably define the class inside the body of the test itself, see the `@action` decorator test for an example (sorry for the spam, commenting on mobile and kept putting it in the wrong place 😛)

NullVoxPopuli: that's what I had initially, but my tests kept failing to start. Not sure how to best reset my environment.

Errors I've gotten:
Cannot find module '../Doc/FileDoc.js'
TypeError: ora is not a function

NullVoxPopuli:
boo, this isn't supposed to pass :-(

NullVoxPopuli:
I'm having a hard time replicating :-(

PR #116, opened 2017-07-05
Fix bug that becomes an invalid type when argument of attr macro is omitted

ursm:
The `@attr` macro has the same type as the property name when the argument is omitted. Therefore, `pushPayload()` on the model causes an error.

``` js
import { attr } from 'ember-decorators/data';

export default Model.extend({
  @attr name: null
});
```

``` js
this.get('store').pushPayload({
  data: {
    type: 'user',
    id: 1,
    attributes: {
      name: 'ursm'
    }
  }
});
//=> Assertion Failed: Unable to find transform for 'name'
```

This patch is modified to handle arguments like normal `DS.attr()`.

### DIFF
--- a/addon/data/index.js
+++ b/addon/data/index.js
@@ -1,4 +1,5 @@
 import DS from 'ember-data';
+import { decoratorWithParams } from '../utils/decorator-wrappers';
 import { decoratorWithKeyReflection } from '../utils/decorator-macros';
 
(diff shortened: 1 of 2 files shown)


rwjblue:
Thank you!

rwjblue:
Published as v1.0.1

PR #117, opened 2017-07-06
feat(accessors): Allows usage of property accessors with the @computed decorator

pzuraq:
Supercedes #61 (thanks for updating the test @offirgolan! Copied them over here 😄)

Removes the guard that prevents users from using the `@computed` decorator on getters and setters. If the descriptor has a getter/setter, the decorator now passes the descriptor directly into the `computedMacro` helper. Not sure if we want to do it this way, or explicitly pull the `get` and `set` functions off the descriptor and create a new object, but this seemed to work pretty well for now.

### DIFF
--- a/addon/object/index.js
+++ b/addon/object/index.js
@@ -147,11 +147,25 @@ export const action = decorator(function(target, key, desc) {
  * @param {...String} propertyNames - List of property keys this computed is dependent on
  */
 export const computed = decoratorWithParams(function(target, key, desc, params) {
-  if (!desc.writable) {
-    throw new Error('ember-decorators does not support using getters and setters');
+  if (desc.value instanceof Ember.ComputedProperty) {
+    throw new Error(`ES6 property getters/setters only need to be decorated once, '${key}' was decorated on both the getter and the setter`);
   }
 
-  let value = extractValue(desc);
+  let value;
+
+  if (desc.writable === undefined) {
+    value = {
+      get: desc.get,
+      set: desc.set
+    };
+
+    // Unset the getter and setter so the descriptor just has a plain value
+    desc.get = undefined;
+    desc.set = undefined;
+  } else {
+    value = extractValue(desc);
+  }
+
   return computedMacro(...params, value);
 });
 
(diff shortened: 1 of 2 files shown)


addon/object/index.js
```diff
@@ -147,11 +147,11 @@ export const action = decorator(function(target, key, desc) {
  * @param {...String} propertyNames - List of property keys this computed is dependent on
  */
 export const computed = decoratorWithParams(function(target, key, desc, params) {
-  if (!desc.writable) {
-    throw new Error('ember-decorators does not support using getters and setters');
+  if (desc.value instanceof Ember.ComputedProperty) {
```
pzuraq: Open to bikeshedding over the correct behavior here. If a user tries to decorate _both_ the getter and setter, the decorator will run twice on the same descriptor (the first time it already has both the getter and setter), so currently we throw an error. 

However, unlike the current `get`/`set` syntax which basically _requires_ both functions to be side-by-side, setters and getters can be anywhere in a class definition. So it may make sense to allow annotation on both to prevent confusion (e.g. this setter _is_ part of a computed, this one is a plain ES setter)

pzuraq: So it turns out this is actually the behavior defined in the decorator spec: http://tc39.github.io/proposal-decorators/#sec-updated-semantics

Step 22: Let coalesced be the new List created by combining getters and setters with the same [[Key]]. If both a getter and setter with the same [[Key]] have a non-empty [[Decorators]], throw a new Error.

So if both the getter _and_ setter are decorated, that's a syntax error, which makes sense.

rwjblue:
> If the descriptor has a getter/setter, the decorator now passes the descriptor directly into the computedMacro helper. 

My concern with this approach is that it seems reasonable for Ember to assert if keys other than `get` and `set` are present on the object passed to `Ember.computed`, and just passing it the raw descriptor would be an issue in that case (it is also quite possible that Ember adds other keys that are not valid in the descriptor that we would have issues setting up).

> explicitly pull the get and set functions off the descriptor and create a new object

Lets swap to this for now.  It seems safer and reserves our ability to change it in the future...

rwjblue:
Other than that change, this looks good to me...

pzuraq:
Done and done

PR #118, opened 2017-07-06
optimization(desc): Modify descriptors instead of creating new ones

### LINKED ISSUE #21: mutate descriptors don't create new descriptors.

pzuraq:
Fixes #21

### DIFF
--- a/addon/utils/decorator-wrappers.js
+++ b/addon/utils/decorator-wrappers.js
@@ -1,12 +1,9 @@
 import isDescriptor from './is-descriptor';
 
 function handleDescriptor(target, key, desc, fn, params = []) {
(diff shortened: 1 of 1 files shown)


pzuraq:
So it looks like the tests are going to fail, it seems that we can't have a property descriptor with `set`/`get` _and_ `value` defined, and setting `set`/`get` to `undefined` doesn't work around that. We can delete the keys instead though, not sure if that would be better than creating a new object.

pzuraq:
cc @stefanpenner @rwjblue

pzuraq:
This will likely change when we update to Babel 7/the new transforms, so closing for now

Issue #89, opened 2017-03-16
@computed causes recalculation of all observed properties


Vladimir85:
By using Ember's computed wrapper, the function is being invoked when at least one of the observable properties has been changed. When recomputing the property, it is not forcing the recalculation of all observables.
```javascript
const { computed, get, isPresent } = Ember;

export default Ember.Object.create({
  firstProperty  : null,
  secondProperty : null,
  
  firstComputed: computed('firstProperty', 'secondComputed', 'thirdComputed', function() {
    if (isPresent(get(this, 'firstProperty'))) {
      // When this condition is verified, "thirdComputed" is not "computed"
      return get(this, 'secondComputed');
    }
    return get(this, 'thirdComputed');
  }),
  
  secondComputed: computed(function() {
    console.log('secondComputed logic executed');
    // some magic here ...
  }),
  
  thirdComputed: computed('otherProperty', function() {
    console.log('thirdComputed logic executed');
    // more magic here ...
  })
});
```
Let's assume that `firstComputed` is used at some point.
On change of the `firstProperty` the `firstComputed` will be recalculated. The `thirdComputed` will be calculated ONLY when the condition `if (isPresent(get(this, 'firstProperty')))` is `false` (exactly as it defined by the logic).

Using `ember-computed-decorators` the behavior of computed property is different.

```javascript
import C from 'ember-computed-decorators';
const { isPresent } = Ember;

export default Ember.Object.create({
  firstProperty  : null,
  secondProperty : null,
  
  @C('firstProperty', 'secondComputed', 'thirdComputed')
  firstComputed(firstProperty, secondComputed, thirdComputed) {
    // At this point all 3 properties in input are 'computed' / 'recalculated'
    if (isPresent(firstProperty)) {
      // When this condition is verified, "thirdComputed" is not "computed"
      return secondComputed;
    }
    return thirdComputed;
  },

  @C('firstProperty', 'secondComputed', 'thirdComputed')
  firstComputedNoParams() {
    // Even at this point all 3 properties in input are 'computed' / 'recalculated'
    if (isPresent(get(this, 'firstProperty'))) {
      // When this condition is verified, "thirdComputed" is not "computed"
      return get(this, 'secondComputed');
    }
    return get(this, 'thirdComputed');
  },

  @C
  secondComputed() {
    console.log('secondComputed logic executed');
    // some magic here ...
  },

  @C('otherProperty')
  thirdComputed(otherProperty) {
[...]

PR #119, opened 2017-07-06
optimization(computed): Use simple computed for accessors

pzuraq:
Opening this partially for discussion. I realized while looking at #89 that using ES6 getters/setters, most of the benefits provided by `ember-macro-helpers` actually don't work at all and only incur extra overhead.

`ember-macro-helpers` relies on being able to have arbitrary function arity. Getters are passed the resolved values for their dependent keys, _or_ the values of previous computeds that have been composed together. Setters also resolve these values, and in both cases that extra data can't be used at all because arity for getter and setter functions is restricted (Trying to make a getter function with any arguments or a setter function with more/less than one is a syntax error in Chrome and a compiler error in Babel.)

I don't think we really have a choice about accepting this behavior which is why I'm proposing this optimization, but it does seem like it could be problematic that the same decorator has somewhat different behavior in different cases.

### DIFF
--- a/addon/object/index.js
+++ b/addon/object/index.js
@@ -1,5 +1,5 @@
 import Ember from 'ember';
-import computedMacro from 'ember-macro-helpers/computed';
+import macroComputed from 'ember-macro-helpers/computed';
 
 import extractValue from '../utils/extract-value';
 import {
@@ -10,6 +10,8 @@ import {
 import {
   decoratorWithRequiredParams
 } from '../utils/decorator-macros';
+
+const { computed: emberComputed } = Ember;
 
 /**
  * Decorator that turns the target function into an Action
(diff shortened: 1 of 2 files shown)


addon/object/index.js
```diff
@@ -151,22 +153,24 @@ export const computed = decoratorWithParams(function(target, key, desc, params)
...
     // Unset the getter and setter so the descriptor just has a plain value
     desc.get = undefined;
     desc.set = undefined;
+
+    // Use a standard ember computed since getter/setter arrity is restricted,
+    // meaning ember-macro-helpers doesn't provide any benefit
+    return emberComputed(...params, {
+      get,
+      set(key, value) {
+        set(value);
```
rwjblue: I think this needs to use some sort of binding to ensure correct `this` (e.g. `.call` / `.apply`).

pzuraq: ah, totally, will update

pzuraq: Updated, and modified a test to make sure it's bound correctly

addon/object/index.js
```diff
@@ -151,22 +153,24 @@ export const computed = decoratorWithParams(function(target, key, desc, params)
...
   }
 
-  let value;
-
   if (desc.writable === undefined) {
-    value = {
-      get: desc.get,
-      set: desc.set
-    };
+    const { get, set } = desc;
```
kellyselden: This should be `let` if you want to stick with Ember style.

pzuraq: I'll defer to @rwjblue on that one, don't really have a preference for style so long as it's consistent

rwjblue: Agreed with @kellyselden.  The rule I'd like to follow is roughly: `const` in module scope or for things that are coming from imports, and `let` for everything else. Essentially "liberal let" from [this blog post](https://madhatted.com/2016/1/25/let-it-be).

pzuraq: :+1: updated! I'll see if I can codify that in eslint and update the rest of the code to match

rwjblue: I believe that emberjs/ember.js has a custom rule for this, maybe look there to see how we're doing it?

Issue #92, opened 2017-05-10
Add helper for addons to bootstrap parent app


thec0keman:
It would be nice if a helper was included with this addon for addons to bootstrap their parent app (described [here](https://github.com/rwjblue/ember-computed-decorators#setup-with-addon)).

This would especially be helpful for some of the API changes with Babel in Ember CLI (e.g. [here](https://github.com/rwjblue/ember-computed-decorators/issues/90)).

Would there be interest in accepting a PR for this?

buschtoens:
Was closed by #121. Addons now handle transpilation themselves independent of the host app.

rwjblue:
Awesome, thanks for spotting that @buschtoens!

PR #121, opened 2017-07-10
enhancement(babel6): Automatically include babel plugins

### LINKED ISSUE #111: Automatically setup host for decorator usage.
I think we can remove https://github.com/rwjblue/ember-decorators#babel-setup and end up with:

```
ember install ember-decorators
```

---

The rough idea would be to add the required configuration on `included` (when running with babel@6).

pzuraq:
fixes #111 and #92 

Adds babel plugins directly to the parent app/addon so users don't have to. We could also add the options for babel 5, but I'm not sure we want continue supporting babel 5 since the spec has already changed and may continue to.


blueprints/ember-decorators/index.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+  normalizeEntityName: function() {}, // no-op since we're just adding dependencies
+
+  afterInstall: function() {
+    return this.addPackagesToProject([
+      { name: 'babel-plugin-transform-class-properties' },
```
rwjblue: We shouldn't need to add these to the project, instead we add them as our own deps and just require them in the babel config (instead of providing a string)

index.js
```diff
@@ -1,6 +1,37 @@
...
+
 module.exports = {
-  name: 'ember-decorators'
+  name: 'ember-decorators',
+
+  included(app) {
+    this._super.included.apply(this, arguments);
+
+    if (!this._registeredWithBabel) {
+      let checker = new VersionChecker(this).for('ember-cli-babel', 'npm');
```
rwjblue: We need to check our parents version of ember-cli-babel (not our own). I think you need to pass `this.parent` not `this`.

blueprints/ember-decorators/index.js
pzuraq: I actually tried doing this at first and it wouldn't build, and I was hesitant about taking control from consumers. Will try again, I'm sure my config was just messed up somehow

index.js
pzuraq: 👍 definitely, makes sense

blueprints/ember-decorators/index.js
rwjblue: Hmm.  Good point. How about we split the difference?  Basically we bring the decorators transform, and we keep the blueprint to add the `transform-class-properties` to the host app...

pzuraq: I think it's ok to include both, the current examples we're laying out require both and as the plugins are updated we can transparently keep it working without users needing to know the details. I think maybe adding a flag to allow users to explicitly opt out may be a better solution, that way we won't throw warning if they're using a different version of babel either.

pzuraq:
@rwjblue is this good to go? Any more changes needed?

index.js
```diff
@@ -1,6 +1,57 @@
...
+}
+
 module.exports = {
-  name: 'ember-decorators'
+  name: 'ember-decorators',
+
+  included(app) {
+    this._super.included.apply(this, arguments);
+
+    const disableTransforms = app.options.emberDecorators && app.options.emberDecorators.disableTransforms;
```
rwjblue: Should use something like https://github.com/babel/ember-cli-babel/blob/master/index.js#L137 to get the options. 

As it is this will not work for nested addons wanting to use decorators.

index.js
```diff
@@ -1,6 +1,57 @@
...
+
+    if (!this._registeredWithBabel && !disableTransforms) {
+      let TransformClassProperties = requireTransform('babel-plugin-transform-class-properties');
+      let TransformDecoratorsLegacy = requireTransform('babel-plugin-transform-decorators-legacy');
+
+      let checker = new VersionChecker(this.parent).for('ember-cli-babel', 'npm');
+
+      if (checker.satisfies('^6.0.0-beta.1')) {
+        app.options = app.options || {};
```
rwjblue: Same as above

rwjblue:
Thank you @pzuraq!

PR #122, opened 2017-07-10
enhancement(debug): Adds no-cost debug helper

pzuraq:
Adds a debug helper which is stripped in production builds, allowing us to assert invariants without adding any runtime cost. The error messages had to be made slightly less descriptive in some cases, but they should be helpful still.

Also refactors some tests to normalize the way we test errors.

### DIFF
--- a/addon/object/computed.js
+++ b/addon/object/computed.js
@@ -27,7 +27,7 @@ import {
  * @function
  * @param {String} dependentKey - Key for the aliased property
  */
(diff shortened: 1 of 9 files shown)


rwjblue:
I'm all for stripping, but we should use the stripping that's baked into ember-cli-babel (and works out of the box). 

Check out:

* https://github.com/babel/ember-cli-babel/blob/master/README.md#debug-tooling
* http://rwjblue.com/2017/05/01/enabling-debug-code-stripping/

PR #126, opened 2017-07-15
bugfix(plugins): Checks plugins with configurations

### LINKED ISSUE #125: Checks for existing plugins should use `name`
These checks https://github.com/rwjblue/ember-decorators/blob/1547d8b7fb928c35b5253c864019384e4c65abe7/index.js#L53 and https://github.com/rwjblue/ember-decorators/blob/1547d8b7fb928c35b5253c864019384e4c65abe7/index.js#L57 should be augmented to look for `name` property on an instance as well as just the plain string value.

We should use something like:

```js
function hasPlugin(plugins, name, instance) {
  for (let plugin of plugins) { // ZOMG we can use let...of with node 4 :tada:
    if (typeof plugin === 'string') {
      if (plugin === name) { return true; }
      continue;
    } else if (Array.isArray(plugin)) {
      if (typeof plugin[0] === 'string')
        if (plugin[0] === name) { return true; }
        continue;
      } else {
        if (plugin[0] === instance) { return true; }
        continue;
      }
    }
  }
  
  return false;
}
```

Although there may be a much better way to write it that that (I just psudeo coded to show the various combos that I am aware of).

Specifically we need to support these scenarios before releasing:

```
// option 1
plugins: [
  'transform-decorators-legacy'
]

// option 2
plugins: [
  ['transform-decorators-legacy']
]

// option 3
plugins: [
  require('babel-plugin-transform-decorators-legacy').default
]
```

pzuraq:
Fixes #125 

Adds a check for plugin name if the plugin has a configuration. Unfortunately it doesn't look like we can check if the plugin instance is being used, see details in #125. Also fixes a bug where the configuration for a parent App wouldn't be found, because `this.parent` will refer to the Project instead of the App.

### DIFF
--- a/index.js
+++ b/index.js
@@ -18,13 +18,29 @@ function requireTransform(transformName) {
   return plugin;
 }
 
+function hasPlugin(plugins, name) {
+  for (let maybePlugin of plugins) {
+    let plugin = Array.isArray(maybePlugin) ? maybePlugin[0] : maybePlugin;
+    let pluginName = typeof plugin === 'string' ? plugin : plugin.name;
+
+    if (pluginName === name) {
+      return true;
+    }
+  }
+
+  return false;
+}
+
 module.exports = {
   name: 'ember-decorators',
 
   _getParentOptions: function() {
     let options;
 
-    if (this.parent) {
+    // The parent can either be an Addon or a Project. If it's an addon,
+    // we want to use the app instead. This public method probably wasn't meant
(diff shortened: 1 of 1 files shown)


pzuraq:
Clarified with @rwjblue that the `name` property will be coming to plugin instances in Babel 7, so added a check against that.

rwjblue:
I mentioned in #125 that I think we can at least try to compare the instances via === check, but I'm happy to add that at a later time if someone actually hits it.

PR #130, opened 2017-07-19
bugfix(dep-check): Updates dependency checker and includes it in dependencies

pzuraq:
`ember-cli-dependency-checker` looks like it didn't get the ability to check parent deps until 2.0.0, and we weren't including it in `dependencies` (big whoops there)

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "babel-plugin-transform-class-properties": "^6.24.1",
     "babel-plugin-transform-decorators-legacy": "^1.3.4",
     "ember-cli-babel": "^6.0.0",
+    "ember-cli-version-checker": "^2.0.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -24,14 +24,14 @@
     "babel-plugin-transform-class-properties": "^6.24.1",
     "babel-plugin-transform-decorators-legacy": "^1.3.4",
     "ember-cli-babel": "^6.0.0",
+    "ember-cli-dependency-checker": "^2.0.0",
```
rwjblue: This is generally only a devDep, how are we using this directly? 

Did you mean ember-cli-version-checker?

pzuraq: Ah, totally confused those two. That explains why it was the wrong version, some other addon I'm using must have included a previous one (I was wondering how that would have happened)

PR #131, opened 2017-07-20
bugfix(actions): Eagerly collapse Ember.Object prototypes

### LINKED ISSUE #128: Actions inherited from non es6 class get overwritten
If you have actions in a parent class that isn't an es6 class declaration and you use the @action decorator it overwrites all of the parent class actions.

pzuraq:
Fixes #128 

Currently the `actions` objects of plain Ember objects written the old-fashioned way doesn't exist at class definition time, it only gets finalized once a class is instantiated. The construction flow can probably be updated to do the linking between the old class style and new class style, but this fix should tide us over (assuming it doesn't cause any nasty side-effects). It calls the `proto` method of the constructor if it exists, eagerly collapsing and creating the `actions` object.

### DIFF
--- a/addon/object/index.js
+++ b/addon/object/index.js
@@ -48,9 +48,18 @@ const { computed: emberComputed } = Ember;
  * @function
  */
 export const action = decorator(function(target, key, desc) {
-  const value = extractValue(desc);
+  let value = extractValue(desc);
 
   assert('The @action decorator must be applied to functions', typeof value === 'function');
+
+  // We must collapse the superclass prototype to make sure that the `actions`
+  // object will exist. Since collapsing doesn't generally happen until a class is
+  // instantiated, we have to do it manually.
+  let superClass = Object.getPrototypeOf(target.constructor);
+
+  if (superClass.hasOwnProperty('proto') && typeof superClass.proto === 'function') {
+    superClass.proto();
+  }
 
   if (!target.hasOwnProperty('actions')) {
     let parentActions = target.actions;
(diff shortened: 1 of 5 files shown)


addon/object/index.js
```diff
@@ -55,8 +55,10 @@ export const action = decorator(function(target, key, desc) {
   // We must collapse the superclass prototype to make sure that the `actions`
   // object will exist. Since collapsing doesn't generally happen until a class is
   // instantiated, we have to do it manually.
-  if (typeof target.constructor.proto === 'function') {
-    target.constructor.proto();
+  let superClass = Object.getPrototypeOf(target.constructor);
+
+  if (superClass.hasOwnProperty('proto') && typeof superClass.proto === 'function') {
```
rwjblue: I think we might want an `in` check? What if we have:

```
let Foo = Ember.Object.extend();
let Bar = class extends Foo {
  @action
  derp() {}
 };
let Baz = class extends Bar {
  @action
  quiz() {}
}
```

Baz should have two actions (derp and qux).

ember-cli-build.js
```diff
@@ -11,6 +11,9 @@ module.exports = function(defaults) {
       config: {
         excludes: ['utils/.*']
       }
+    },
+    'ember-cli-babel': {
+      includePolyfill: true
```
rwjblue: I don't think we need this for Object.getPrototypeOf...

pzuraq: This was for regenerator/async-await style tests, `Promise` doesn't exist in PhantomJS yet (trolled me so hard 😛)

addon/object/index.js
pzuraq: Definitely will add a test, but I do think this works as is. Once we hit the ES class chain, all the classes get eagerly evaluated, so actions should exist by the time the next class is being defined

PR #137, opened 2017-07-25
docs: ES6 classes and RFC 176 module imports (#120)

### LINKED ISSUE #70: ESLint undefined errors
How can I get ESLint to not flag things like this as undefined?

`@alias('user.address.validations.attrs.state.messages') stateErrorMessages,`

`error  'stateErrorMessages' is not defined  no-undef`

### LINKED ISSUE #120: Update documentation to showcase class syntax.
Need to update the following files to showcase native ES `class` syntax primarily with a secondary example showing the existing `.extend()` syntax (and including the `@decorator foo: null` work around as required).

- [ ] `ember-decorators/controller`
- [ ] `ember-decorators/data`
- [ ] `ember-decorators/object`
- [ ] `ember-decorators/object/computed`
- [ ] `ember-decorators/object/evented`
- [ ] `ember-decorators/service`

buschtoens:
Closes #120.
Closes #70, in that it only features ES6 class syntax or valid (`@attr prop: null`) EmberObject model syntax.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ import Component from '@ember/component'
 import { action, computed } from 'ember-decorators/object';
 import { service } from 'ember-decorators/service';
 
-export default class extends Component {
+export default class ExampleComponent extends Component {
   @service foo
 
   @computed('someKey', 'otherKey')
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ export default Ember.Component.extend({
     var someKey = this.get('someKey');
     var otherKey = this.get('otherKey');
 
-    // do stuff
+    return `${someKey} - ${otherKey}`;
   }),
 
   actions: {
@@ -46,16 +46,16 @@ export default Ember.Component.extend({
 You replace it with this:
 
 ```javascript
-import Ember from 'ember'
+import Component from '@ember/component'
 import { action, computed } from 'ember-decorators/object';
 import { service } from 'ember-decorators/service';
 
-export default class MyComponent extends Ember.Component {
+export default class extends Component {
   @service foo
 
   @computed('someKey', 'otherKey')
   bar(someKey, otherKey) {
-    // do stuff
+    return `${someKey} - ${otherKey}`;
   }
 
   @action
--- a/addon/data/index.js
+++ b/addon/data/index.js
@@ -9,9 +9,9 @@ import { decoratorWithKeyReflection } from '../utils/decorator-macros';
  * import Model from 'ember-data/model';
  * import { attr } from 'ember-decorators/data';
  *
- * export default class extends Model {
- *   @attr firstName;
- * }
+ * export default Model.extend({
+ *   @attr firstName: null
+ * });
  * ```
  *
  * @function
@@ -28,9 +28,9 @@ export const attr = decoratorWithParams(function(target, key, desc, params) {
  * import Model from 'ember-data/model';
  * import { hasMany } from 'ember-decorators/data';
  *
- * export default class extends Model {
- *   @hasMany users;
- * }
+ * export default Model.extend({
+ *   @hasMany users: null
+ * });
  * ```
  *
  * @function
@@ -45,9 +45,9 @@ export const hasMany = decoratorWithKeyReflection(DS.hasMany);
  * import Model from 'ember-data/model';
  * import { belongsTo } from 'ember-decorators/data';
  *
- * export default class extends Model {
- *   @belongsTo user;
- * }
+ * export default Model.extend({
+ *   @belongsTo user: null
+ * });
  * ```
  * @function
  * @param {String} [type] - Type of the `belongsTo` relationship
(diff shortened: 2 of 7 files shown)


buschtoens:
Whoops. Don't merge yet. I screwed up the class properties.

buschtoens:
Should be all good now. :grinning:

addon/object/index.js
```diff
@@ -74,82 +65,86 @@ export const action = decorator(function(target, key, desc) {
...
  *     set(value, first, last) {
- *       // ...
+ *       if (typeof value !== 'string' || !value.test(/^[a-z]+ [a-z]+$/i)) {
+ *         throw new TypeError('Invalid name');
+ *       }
+ *
+ *       const [first, last] = value.split(' ');
+ *       setProperties(this, { first, last });
+ *
+ *       return value;
```
buschtoens: In the spirit of slimming down the docs, this example might be a bit over the top, however I find it important to at least include the `return value;` statement. Otherwise users might easily trip over this subtle bug.

pzuraq: I think that's reasonable, better to be clear than concise

pzuraq: I think that's reasonable, better to be clear than concise

README.md
```diff
@@ -46,16 +46,16 @@ export default Ember.Component.extend({
 You replace it with this:
 
 ```javascript
-import Ember from 'ember'
+import Component from '@ember/component'
 import { action, computed } from 'ember-decorators/object';
 import { service } from 'ember-decorators/service';
 
-export default class MyComponent extends Ember.Component {
```
pzuraq: I think we should encourage users to name their classes, it's one of the benefits of ES Classes (easier to grep for class names, no need for a class to be resolved to know its type - this plagues us with classes we use that are not resolved)

addon/data/index.js
```diff
@@ -6,12 +6,12 @@ import { decoratorWithKeyReflection } from '../utils/decorator-macros';
...
  *
  * ```javascript
- * import DS from 'ember-data';
+ * import Model from 'ember-data/model';
  * import { attr } from 'ember-decorators/data';
  *
- * export default DS.Model.extend({
- *   @attr firstName
- * });
+ * export default class extends Model {
```
buschtoens: Potentially harmful: https://github.com/rwjblue/ember-decorators/issues/98#issuecomment-312514529

pzuraq: Agreed, we should keep these examples as normal Ember Objects until we iron out their usage

README.md
buschtoens: Do you know of any plans on making the default blueprints do that as well? Seeing that Module Unification apps will disallow nested components anyways (which would have yielded absurdly long class names), I think this is a nice feature.

In that case, I think we should use "valid" class names in our examples, like `UserProfileComponent` (to fulfill the "include a dash" constraint) instead of `MyComponent`.

pzuraq: Also, easier to document - ESDoc automatically generates documentation for classes, it would be difficult to do it well for "anonymous" classes

addon/object/index.js
rwjblue: I'm not sure that this actually works?  See [here](https://github.com/rwjblue/ember-decorators/blob/master/addon/object/index.js#L173) we do not `return set.call(......)`.

README.md
rwjblue: Confirm. It also makes debugging nicer IMHO...

addon/data/index.js
rwjblue: Agreed

addon/object/index.js
buschtoens: Good catch.

Since `emberComputed` is [`Ember.computed`](https://emberjs.com/api/ember/2.14/namespaces/Ember.computed), I would argue that this is broke at the moment.

rwjblue: ya, lets leave these docs with the return call (since it should work) and make another issue to track fixing setter CP's (and beef up some tests).

README.md
pzuraq: We should consider adding some blueprints that generate the new style and would add the classname. Definitely agree that we should use realistic class names

addon/object/index.js
pzuraq: Oh yeah, definitely a bug. Should note that this particular example `macroComputed` will be used instead actually, since it's an object that's a class property and not class getters/setters, but class setters are almost definitely broken.

rwjblue:
Looks like this needs a rebase now that the other PR's have landed.

rwjblue:
Looks like I was able to resolve the conflict via the GH web UI.

rwjblue:
@pzuraq - I think all of our prior concerns have been addressed, mind double checking (I don't want to merge if you still have open concerns...)?

buschtoens:
The docs that were merged in from master were written in the EmberObject model context. I'll send a PR tomorrow.

PR #139, opened 2017-07-25
feat(object/computed): filter, map, sort: fallback to descriptor value for callback (#138)

### LINKED ISSUE #138: filter, map, sort: use descriptor value as callback
Currently [`filter(dependentKey, callback)`](https://rwjblue.github.io/ember-decorators/docs/function/addon/object/computed.js.html#static-function-filter), [`map(dependentKey, callback)`](https://rwjblue.github.io/ember-decorators/docs/function/addon/object/computed.js.html#static-function-map) and [`sort(dependentKey, sortDefinition)`](https://rwjblue.github.io/ember-decorators/docs/function/addon/object/computed.js.html#static-function-sort) all expect the callback function to be passed as the last argument, just like their `Ember.computed` counterparts. This makes for (subjectively) ugly syntax and nagged me since day one.

I propose that we make the callback argument optional (for these three decorators only) and fallback to the descriptor value. Just like we normally do for `@computed`.

This would allow this syntax:

```js
import EmberObject from '@ember/object';
import { A } from '@ember/array';
import {filter, sort} from 'ember-decorators/object/computed'

export default class extends EmberObject {
  chores = A([
    { name: 'cook', done: true },
    { name: 'clean', done: true },
    { name: 'write more unit tests', done: false }
  ]);

  @filter('chores')
  remainingChores(chore, index, array) {
    return !chore.done;
  }

  @map('chores')
  loudChores(chore, index) {
    return chore.name.toUpperCase() + '!';
  }

  @sort('names')
  sortedNames(a, b) {
    if (a.name > b.name) {
      return 1;
    } else if (a.name < b.name) {
      return -1;
    }

    return 0;
  }
}
```

Only falling back to the descriptor value when the callback argument is empty has two advantages:

- full backwards compatibility
- hypothetical composability still possible (#99)

Yay or nay? I can try to submit a PR. :slightly_smiling_face:

buschtoens:
Also adds the `decoratedPropertyWithOptionalCallback` decorator macro.
    
Closes #138.

Needs rebase, when #137 is merged, or the other way around.

### DIFF
--- a/addon/utils/decorator-macros.js
+++ b/addon/utils/decorator-macros.js
@@ -39,10 +39,8 @@ export function decoratedPropertyWithOptionalCallback(fn) {
     }
 
     const value = extractValue(desc);
-    if (typeof value === 'function') {
-      return fn(...params, value);
-    }
+    assert(`Cannot use '${fn.name}' on field '${key}' without a callback`, typeof value === 'function');
 
-    assert(`Cannot use '${fn.name}' on field '${key}' without a callback`, false);
+    return fn(...params, value);
   });
 }
--- a/addon/utils/decorator-macros.js
+++ b/addon/utils/decorator-macros.js
@@ -29,3 +29,20 @@ export function decoratedPropertyWithRequiredParams(fn) {
     return fn(...params);
   });
 }
+
+export function decoratedPropertyWithOptionalCallback(fn) {
+  return decoratorWithParams(function(target, key, desc, params) {
+    assert(`Cannot use '${fn.name}' on field '${key}' without parameters`, params.length !== 0);
+
+    if (typeof params[params.length - 1] === 'function') {
+      return fn(...params);
+    }
+
+    const value = extractValue(desc);
+    if (typeof value === 'function') {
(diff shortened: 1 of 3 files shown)


addon/utils/decorator-macros.js
```diff
@@ -29,3 +29,20 @@ export function decoratedPropertyWithRequiredParams(fn) {
...
+export function decoratedPropertyWithOptionalCallback(fn) {
+  return decoratorWithParams(function(target, key, desc, params) {
+    assert(`Cannot use '${fn.name}' on field '${key}' without parameters`, params.length !== 0);
+
+    if (typeof params[params.length - 1] === 'function') {
+      return fn(...params);
+    }
+
+    const value = extractValue(desc);
+    if (typeof value === 'function') {
```
pzuraq: Could probably do this check eagerly in the `assert` below so that it gets stripped in production

pzuraq:
Looks good aside from that small optimization

pzuraq:
👍 thanks for knocking this out @buschtoens!

PR #141, opened 2017-07-25
fix(object): computed: require getter, don't require but honor setter return value

### LINKED ISSUE #140: @computed: disallow lonely setter, ignore its return value, always use getter return value as property value
https://github.com/rwjblue/ember-decorators/pull/137#discussion_r129347557

Apparently setting a `@computed` property with a setter will always set that CP to `undefined` [since the return value is not passed on](https://github.com/rwjblue/ember-decorators/blob/8a36530c238e823ce9bce8259bf604606aa8ecfa/addon/object/index.js#L173).

buschtoens:
Fixes #140.

### DIFF
--- a/addon/object/index.js
+++ b/addon/object/index.js
@@ -1,3 +1,5 @@
+import { DEBUG } from '@glimmer/env';
+
 import Ember from 'ember';
 import macroComputed from 'ember-macro-helpers/computed';
 
@@ -9,7 +11,7 @@ import {
 
 import { decoratorWithRequiredParams } from '../utils/decorator-macros';
 
-import { assert, warn } from '@ember/debug';
+import { assert } from '@ember/debug';
 
 const { computed: emberComputed } = Ember;
 
@@ -167,21 +169,22 @@ export const computed = decoratorWithParams(function(target, key, desc, params) 
     desc.get = undefined;
     desc.set = undefined;
 
+    let setter;
+
+    if (typeof set === 'function') {
+      setter = function(key, value) {
+        let ret = set.call(this, value);
+        return typeof ret === 'undefined' ? get.call(this) : ret;
+      };
+    } else if (DEBUG) {
+      setter = function() {
+        assert(`You must provide a setter in order to set '${key}' as a computed property.`, false);
+      };
+    }
+
     // Use a standard ember computed since getter/setter arrity is restricted,
     // meaning ember-macro-helpers doesn't provide any benefit
-    return emberComputed(...params, {
-      get,
-      set(key, value) {
-        const ret = set.call(this, value);
-        warn(
-          `The return value ('${ret}') of the ES6 setter for '${key}' was ignored. For more information, see: https://github.com/rwjblue/ember-decorators/pull/141#issuecomment-317974858`,
-          typeof ret === 'undefined',
-          { id: 'ember-decorators.object.computed.es6-setter-return-value' }
-        );
-
-        return get.call(this);
-      }
-    });
+    return emberComputed(...params, { get, set: setter });
   } else {
     return macroComputed(...params, extractValue(desc));
   }
--- a/addon/object/index.js
+++ b/addon/object/index.js
@@ -170,7 +170,7 @@ export const computed = decoratorWithParams(function(target, key, desc, params) 
     return emberComputed(...params, {
       get,
       set(key, value) {
-        set.call(this, value);
+        return set.call(this, value);
       }
     });
   } else {
--- a/addon/object/index.js
+++ b/addon/object/index.js
@@ -9,7 +9,7 @@ import {
 
 import { decoratorWithRequiredParams } from '../utils/decorator-macros';
 
-import { assert } from '@ember/debug';
+import { assert, warn } from '@ember/debug';
 
 const { computed: emberComputed } = Ember;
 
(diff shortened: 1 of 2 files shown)


rwjblue:
I wonder if we can take this opportunity to remove the wackiness here. Specifically, we could do something like:

```js
    return emberComputed(...params, {
      get,
      set(key, value) {
        let ret = set.call(this, value);
        if (ret === undefined) {
          return get.call(this);
        }
      }
    });
```

buschtoens:
Assuming that users will never intentionally `return undefined;`, this seems like a good idea.

buschtoens:
However, this would only work for ES6 getters and setters then. I think we should rather be consistent.

rwjblue:
We are already in an uncanny valley (using getters and setters opts you into a completely different world than using methods), and I think we will have to dig out via deprecations and a major version bump...

Specifically, I think we should drop support for the "method" version:

```js
class Foo extends Ember.Object {
  @computed
  derp() {}
}
```

And only support the getter and/or setter versions:

```js
class Foo extends Ember.Object {
  @computed
  get derp() {}
}
```

If we do this, then my proposed change is not harmful or inconsistent (there would just be one way to do it). 

---

Also, FWIW, we want to make this change in Ember itself also.  Basically, deprecating returning `undefined` for a given set when the getter for the CP returns something other than `undefined`.

rwjblue:
> Assuming that users will never intentionally return undefined;, this seems like a good idea.

Even if they do `return undefined` on purpose, it is still ok as long as their getter matches (and we can add a warning/deprecation/assertion/etc if they don't).

pzuraq:
This _is_ what `null` is for after all 😛 I think it makes sense. That major version bump is going to be coming either way, and I agree that getters/setters seem like the path forward, so we can experiment a little with this I think.

buschtoens:
Ah okay. I didn't know about these plans. Makes sense.

I'll add it to the PR.

buschtoens:
Would you like a warning when a setter returns `undefined` and there is no corresponding getter?

pzuraq:
I think a warning would make sense, I can't think of a case where you would want to use a computed `setter` without a `getter` over just a plain ES setter. Like, it could clear out the cached value whenever a key updates? I think there would be better ways to accomplish that.

Actually thinking this through more, we should probably mirror the behavior of ES getters and setters and throw if a user tries to get without a getter defined or set without a setter defined.

buschtoens:
Thinking about it, since the return value of an [ES6 setter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/set) is ignored and we force users to have a getter as well, we should also ignore the return value and just return the getter's value, i.e.:

```js
    return emberComputed(...params, {
      get,
      set(key, value) {
        const ret = set.call(this, value);
        warn(`Return values of ES6 setters are ignored.`, typeof ret === 'undefined');
        return get.call(this);
      }
    });
```

pzuraq:
I'm on board with requiring a getter for computeds, there are no use cases I can think of where a plain ES6 setter without a getter makes sense as a computed. _Technically_ you could do:

```javascript
const Foo = Ember.Object.extend({
  bar: computed({
    set(key, val) {
      return val;
    }
  })
});

foo.set('val', 1);
console.log(foo.get('val')); // 1
```

But I think that's horribly, terribly awful behavior that we shouldn't enable.

I also agree that we should enable ES6 setter semantics - essentially, if you have an ES6 getter + setter and you want to mark them as a computed, you should be able to do so with no changes. However, I think we should consider providing an opt-out by still accepting the return value if it exists. This makes sense since it would essentially be providing additional functionality on top of ES6 setters, which decorators are able to do. So:

```javascript
set (key, value) {
  assert(`you must provide a setter in order to set '${key}' as a computed property`, typeof set === 'function');

  let ret = set.call(this, value);
  return ret === undefined ? get.call(this) : ret;
}
```

Added an assertion, the code already implies that `set` must exist but this will give users a nicer error message.

pzuraq:
@rwjblue thoughts?

pzuraq:
we can also probably eliminate the need for creating a setter function if one does not exist:

```
import { DEBUG } from '@glimmer/env';

...

let { get, set } = desc;

assert(`Using @computed for only a setter does not make sense. Add a getter for '${key}' as well or remove the @computed decorator.`, typeof get === 'function');

// Unset the getter and setter so the descriptor just has a plain value
desc.get = undefined;
desc.set = undefined;

let setter;

if (typeof set === 'function') {
  setter = function(key, value) {
    let ret = set.call(this, value);
    return ret === undefined ? get.call(this) : ret;
  };
} else if (DEBUG) {
  setter = function() {
    assert(`you must provide a setter in order to set '${key}' as a computed property`, typeof set === 'function');
  };
}

// Use a standard ember computed since getter/setter arrity is restricted,
// meaning ember-macro-helpers doesn't provide any benefit
return emberComputed(...params, { get, set: setter });
```

rwjblue:
Agreed, https://github.com/rwjblue/ember-decorators/pull/141#issuecomment-318022389 seems like the right way forward to me also.

pzuraq:
👍 Looks good! Thanks for getting this over the line @buschtoens!

buschtoens:
Glad I could help. :smiley:

PR #143, opened 2017-07-25
feat(object/computed): add deprecatingAlias

buschtoens:
Closes #127.

This does not (yet) include the babel transform suggested in https://github.com/rwjblue/ember-decorators/issues/127#issuecomment-315808212. My knowledge of Babel transforms is limited, so maybe @Cryrivers would like to step in here.

PR #148, opened 2017-07-27
feat(object/computed): add uniqBy

buschtoens:
Docs: http://emberjs.com/api/classes/Ember.computed.html#method_uniqBy

Closes #71.

I guess this one must have fallen through the cracks. Together with #143 we have full feature parity with `Ember.computed`. :tada: :grinning:

### DIFF
--- a/addon/object/computed.js
+++ b/addon/object/computed.js
@@ -1,10 +1,9 @@
 import Ember from 'ember';
+import { assert } from '@ember/debug';
 import {
   decoratedPropertyWithRequiredParams,
   decoratedPropertyWithOptionalCallback
 } from '../utils/decorator-macros';
-
-const { assert } = Ember;
 
 /**
  * Decorator that wraps [Ember.computed.alias](http://emberjs.com/api/classes/Ember.computed.html#method_alias)
@@ -784,4 +783,6 @@ export const uniq = decoratedPropertyWithRequiredParams(Ember.computed.uniq);
  */
 export const uniqBy = Ember.computed.uniqBy ?
   decoratedPropertyWithRequiredParams(Ember.computed.uniqBy) :
-  () => assert('uniqBy is only available from Ember.js v2.7 onwards.', false);
+  () => {
+    assert('uniqBy is only available from Ember.js v2.7 onwards.', false);
+  };
--- a/addon/object/computed.js
+++ b/addon/object/computed.js
@@ -1,4 +1,5 @@
 import Ember from 'ember';
+import { assert } from '@ember/debug';
 
 import {
   decoratedPropertyWithRequiredParams,
@@ -781,4 +782,6 @@ export const uniq = decoratedPropertyWithRequiredParams(Ember.computed.uniq);
  * @param {String} dependentKey - Key of the array to uniq
  * @param {String} propertyKey - Key of the property on the objects of the array to determine uniqueness by
  */
-export const uniqBy = decoratedPropertyWithRequiredParams(Ember.computed.uniqBy);
+export const uniqBy = Ember.computed.uniqBy ?
+  decoratedPropertyWithRequiredParams(Ember.computed.uniqBy) :
+  () => assert('uniqBy is only available from Ember.js v2.7 onwards.', false);
--- a/addon/object/computed.js
+++ b/addon/object/computed.js
@@ -1,10 +1,10 @@
 import Ember from 'ember';
-import { assert } from '@ember/debug';
-
 import {
   decoratedPropertyWithRequiredParams,
   decoratedPropertyWithOptionalCallback
 } from '../utils/decorator-macros';
+
+const { assert } = Ember;
 
 /**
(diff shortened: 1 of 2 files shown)


rwjblue:
I believe that uniqBy was added relatively recently (in 2.5 or 2.6 IIRC), which is why CI for Ember 2.4 is failing. We probably need to guard the test so that it doesn't run when not supported...

buschtoens:
Oh yeah. Makes sense. Would you like me to guard the test, or should I rather guard the helper itself, so that it throws a descriptive error, when used?

rwjblue:
@buschtoens - I'd like both  😝 .

So, in the test:

```js
import hasEmberVersion from 'ember-test-helpers/has-ember-version';

// ....snip....

if (hasEmberVersion(2,6)) {
  test('uniqBy .....');
}
```

And in the implementation:

```js
if (!Ember.computed.uniqBy) {
  assert(...);
}
```

buschtoens:
Turns out `uniqBy` was added in Ember v2.7: [Blog post](https://www.emberjs.com/blog/2016/07/25/ember-2-7-and-2-8-beta-released.html#toc_code-ember-computed-uniqby-code)

buschtoens:
Dammit.. I can't get `ember-try` to run in `serve` mode for some reason...

```
ember try:one ember-lts-2.4 --- ember serve
=== Scenario: ember-lts-2.4 ====================================================

Error!
The command ember serve exited Error: Cannot find module 'ember-cli' from '/home/jan/github/ember-decorators'
undefined
```

I'll debug it somehow else. :sweat_smile:

kellyselden:
I had this same problem [here](https://github.com/kellyselden/ember-awesome-macros/issues/269). It would be nice if we could combine efforts and make a true polyfill addon we both could share.

rwjblue:
@buschtoens - The issue in https://github.com/rwjblue/ember-decorators/commit/41b89a57f11dffcf6f9e3ce50a39965740223617 is what I described in https://github.com/babel/ember-cli-babel/issues/168#issuecomment-318042918 and opened https://github.com/chadhietala/babel-plugin-debug-macros/issues/46 about.

The fix for that commit is to not `return assert(...)`:

```js
export const uniqBy = Ember.computed.uniqBy ?
  decoratedPropertyWithRequiredParams(Ember.computed.uniqBy) :
  () => { assert('uniqBy is only available from Ember.js v2.7 onwards.', false); };
```

addon/object/computed.js
```diff
@@ -1,10 +1,11 @@
 import Ember from 'ember';
-
 import {
   decoratedPropertyWithRequiredParams,
   decoratedPropertyWithOptionalCallback
 } from '../utils/decorator-macros';
 
+const { assert } = Ember;
```
rwjblue: Sorry for the issues in the underlying lib, we definitely want to use `import { assert } from '@ember/debug';` though. See https://github.com/rwjblue/ember-decorators/pull/148#issuecomment-318432401 for a work around.

buschtoens: No worries. 😊

While we're at it, would you like another PR upgrading the rest of the code base to use module imports?

rwjblue: sure!

pzuraq: My only concern with module imports is backwards compat, we're using `ember-decorators` on a fairly old version of Ember/ember-cli-shims

rwjblue: @pzuraq - There is no ember-cli-shims version requirement. As long as ember-decorators depends on `ember-cli-babel@^6.6.0` we get [the polyfill](https://github.com/ember-cli/babel-plugin-ember-modules-api-polyfill) for free.  This means that

```js
import { inject } from "@ember/service"
```

is transpiled back to 

```js
const inject = Ember.inject.service;
```

Which is basically what we are doing today (with the destructuring).  This change simply lets this addon prepare for consuming Ember in a way that will allow actual tree shaking down the line...

pzuraq: Ah, that's awesome! Cool, I'll be converting us internally then as well 😁

rwjblue: You can use the lovely [ember-modules-codemod](https://github.com/ember-cli/ember-modules-codemod) to make the change quickly 😸

rwjblue:
Thank you!

kellyselden:
Anyone have any comments on https://github.com/rwjblue/ember-decorators/pull/148#issuecomment-318419940?

pzuraq:
I think a polyfill would be great, working on a general purpose flag addon that'll allow us to check that these features either exist in the current version of Ember or have been polyfilled: https://github.com/pzuraq/ember-compatibility-helpers

buschtoens:
I would also be in favor. Do you think it'll be possible to conditionally include the polyfill only for versions that need it?

pzuraq:
yes, checkout @rwjblue's work on the [factoryFor polyfill](https://github.com/rwjblue/ember-factory-for-polyfill/blob/master/index.js#L17-L19), should be possible to do something similar

Issue #127, opened 2017-07-17
Request: deprecatingAlias


buschtoens:
[API docs for `Ember.computed.deprecatingAlias`](https://www.emberjs.com/api/ember/2.14/namespaces/Ember.computed/methods/deprecatingAlias?anchor=deprecatingAlias)

Usage:

```diff
import Component from '@ember/component';
-import { deprecatingAlias } from '@ember/object/computed';
+import { deprecatingAlias } from 'ember-decorators/object/computed';

export default class extends Component {
  newProperty;

-  deprecatedProperty = deprecatingAlias('newProperty', {
-    id: 'foo-deprecation',
-    until: '1.0.0',
-    link: 'https://example.com/deprecations/foo-deprecation
-  });
+  @deprecatingAlias('newProperty', {
+    id: 'foo-deprecation',
+    until: '1.0.0',
+    link: 'https://example.com/deprecations/foo-deprecation
+  })
+  deprecatedProperty;
}
```

I'd submit a PR, if this is a welcome addition. :slightly_smiling_face:

pzuraq:
I think this would be a good addition. I would like to strip it from production builds with a babel transform and replace it with a normal `@alias` decorator so we avoid creating the options object and the extra overhead, but given the purpose of the decorator I think we can go with the simple implementation for now and add the transform to remove it later on.

Cryrivers:
@pzuraq i can help it (decorator + babel transformer)

rwjblue:
We can use ember-cli-babel's built in `import { DEBUG } from '@glimmer/env'` to detect if we are in a debug build, and change which underlying implementation to use. 

:+1: on this

PR #149, opened 2017-07-28
WIP: Transform deprecatingAlias to alias at compile time

Cryrivers:
Optimizations for #127.

### DIFF
--- a/lib/babel-plugins.js
+++ b/lib/babel-plugins.js
@@ -1,0 +1,44 @@
+export default function (babel) {
+  
+  const { types: t } = babel;
+  
+  function deprecateAliasToAlias(path) {
+    const { decorators } = path.node;
+    const updatedDecorators = decorators.map(decorator => {
+      if (decorator.expression && 
+          decorator.expression.type === 'CallExpression' &&
+          decorator.expression.callee.name === 'deprecatingAlias') {
+        return t.decorator(
+          t.callExpression(
+            t.Identifier('alias'),
+            [decorator.expression.arguments[0]]
+          )
+        );
+      }
+      return decorator;
+    });
+    path.node.decorators = updatedDecorators;
+  }
+  
+  return {
+    visitor: {
+      ImportDeclaration(path) {
+        if (path.node.source.value === 'ember-decorators/object/computed') {
+          path.node.specifiers = path.node.specifiers.map(specifier => {
+            if (specifier.imported.name === 'deprecatingAlias') {
+              return t.importSpecifier(t.Identifier('alias'), t.Identifier('alias'));
+            } else {
+              return specifier;
+            }
+          });
+        }
+      },
+      ClassProperty(path) {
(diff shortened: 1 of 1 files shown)


buschtoens:
If I understand correctly, this transform does the following:

1) replace `deprecatingAlias` imports with `alias`, i.e. turn `import { deprecatingAlias } from 'ember-decorators/object/computed';` into `import { alias } from 'ember-decorators/object/computed';`
2) replace all class and object `@deprecatingAlias` decorators with `@alias`

Wouldn't **1** cause problems, when the user already imports `alias`?

```js
import { deprecatingAlias, alias } from 'ember-decorators/object/computed';
// ... would become:
import { alias, alias } from 'ember-decorators/object/computed';
```
Instead of changing all object and class decorators as well, wouldn't it be easier and safer to only modify the import, so that all `deprecatingAlias` get changed to `alias as deprecatingAlias`?

```js
import {
  deprecatingAlias,
  alias,
  deprecatingAlias as secretDeprecatingAlias
} from 'ember-decorators/object/computed';
// ... would become:
import {
  alias as deprecatingAlias,
  alias,
  alias as secretDeprecatingAlias
} from 'ember-decorators/object/computed';
```

That they, there won't be any import collisions and the transform is a bit faster.

Cryrivers:
Great suggestion! This PR is still WIP. The code I committed is the result I played around with ast-explorer. I’ll try to cover all cases this weekend.

buschtoens:
I myself have absolutely no idea of babel, so I'm glad you're taking this on. :smile:

pzuraq:
This is great, but I think we can probably use the DEBUG helper included in ember-cli-babel by default now: https://github.com/babel/ember-cli-babel#general-purpose-env-flags

The idea is that in production, the DEBUG flag will be replaced with a boolean literal `false`, so

```
if (DEBUG) { ... }
```

becomes:

```
if (false) { ... }
```

And minifiers can then use dead-code elimination to remove the statement entirely.

buschtoens:
So basically this?

```js
import { DEBUG } from '@glimmer/env';
import {
  alias,
  deprecatingAlias as _deprecatingAlias
} from 'ember-decorators/object/computed';
const deprecatingAlias = DEBUG ? _deprecatingAlias : alias;
```

pzuraq:
Yes, pretty sure that would work (not positive though, might be good to double check)

rwjblue:
Ya, that should work for the swapping the underlying implementation but it doesn't remove the cost of the options in the host app/addon itself (e.g. if calculating the message was expensive).

Cryrivers:
yeah i think @pzuraq 's method is also feasible. However, I'm not quite sure when to insert `import { DEBUG } from '@glimmer/env'; `.

pzuraq:
You should probably import it at the top of the file, with the other import statements. I think it's technically not allowed to import anything after other statements have occured

PR #150, opened 2017-07-29
enhancement(compat): Ensure backwards compatibility to 1.11, adds compat helpers

pzuraq:
Adds ember-try scenarios for 1.11 and 1.13, and adds `ember-compatibility-helpers` which gives us build time flags for Ember feature detection. This means we can write forward/backwards compatible code with absolutely no dead weight added to production builds of consuming applications.

@rwjblue I included the fix for getters/setters, if you find time to update the polyfill I'll update the compat helpers and remove the fallback

### DIFF
--- a/addon/object/index.js
+++ b/addon/object/index.js
@@ -12,6 +12,10 @@ import {
 import { decoratorWithRequiredParams } from '../utils/decorator-macros';
 
 import { assert } from '@ember/debug';
+import {
+  HAS_UNDERSCORE_ACTIONS,
+  SUPPORTS_NEW_COMPUTED
+} from 'ember-compatibility-helpers';
 
 const { computed: emberComputed } = Ember;
 
@@ -54,12 +58,21 @@ export const action = decorator(function(target, key, desc) {
     superClass.proto();
   }
 
-  if (!target.hasOwnProperty('actions')) {
-    let parentActions = target.actions;
-    target.actions = parentActions ? Object.create(parentActions) : {};
-  }
+  if (HAS_UNDERSCORE_ACTIONS) {
+    if (!target.hasOwnProperty('_actions')) {
+      let parentActions = target._actions;
+      target._actions = parentActions ? Object.create(parentActions) : {};
+    }
 
-  target.actions[key] = value;
+    target._actions[key] = value;
+  } else {
+    if (!target.hasOwnProperty('actions')) {
+      let parentActions = target.actions;
+      target.actions = parentActions ? Object.create(parentActions) : {};
+    }
+
+    target.actions[key] = value;
+  }
 
   return value;
 });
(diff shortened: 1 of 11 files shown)


addon/object/index.js
```diff
@@ -179,7 +191,17 @@ export const computed = decoratorWithParams(function(target, key, desc, params)
 
     // Use a standard ember computed since getter/setter arrity is restricted,
     // meaning ember-macro-helpers doesn't provide any benefit
-    return emberComputed(...params, { get, set: setter });
+    if (SUPPORTS_NEW_COMPUTED) {
+      return emberComputed(...params, { get, set: setter });
+    } else {
+      return emberComputed(...params, function() {
```
rwjblue: I don't think this will work, will it? You need to make the function arity > 1 or Ember won't use the setter (and instead will clobber), right? Am I completely mis-remembering this?

pzuraq: could be true, I haven't used this style in a long time 😛 I'll update it to have args, no reason to choose one or the other

pzuraq:
Looks like failures on canary, beta, and release. Can ember-cli-version-checker handle those cases? We can add logic in ember-compatibility-helpers if not

addon/object/index.js
rwjblue: Im thinking of this code: https://github.com/emberjs/ember.js/blob/v1.11.0/packages/ember-metal/lib/computed.js#L121-L123

rwjblue:
No, I think ember-cli-version-checker may not be able to properly identify beta/canary/etc.

pzuraq:
mmk, looks like 1.11 might be failing for other reasons (weird, thought I'd run Ember try on it before). I'll dig in, and get those sample outputs.

rwjblue:
Awesome.  Excited to get this under test.

pzuraq:
Tests fixed, ready to review!

addon/object/index.js
```diff
@@ -179,7 +192,17 @@ export const computed = decoratorWithParams(function(target, key, desc, params)
 
     // Use a standard ember computed since getter/setter arrity is restricted,
     // meaning ember-macro-helpers doesn't provide any benefit
-    return emberComputed(...params, { get, set: setter });
+    if (SUPPORTS_NEW_COMPUTED) {
+      return emberComputed(...params, { get, set: setter });
+    } else {
+      return emberComputed(...params, function(key, value) {
```
rwjblue: We need a conditional here I believe. Otherwise it's not possible to clobber (we will always be treated like a setter was specified).

Something like:

```
let callback;
if (setter) {
  callback = function (key, value) {
    if (arguments.length > 1) {
      return setter.call(this, key, value);
    }

    return get.call(this);
  }
} else {
  callback = get;
}

return emberComputed(...params, callback);
```

package.json
```diff
@@ -25,6 +25,7 @@
     "babel-plugin-transform-decorators-legacy": "^1.3.4",
     "ember-cli-babel": "^6.0.0",
     "ember-cli-version-checker": "^2.0.0",
+    "ember-compatibility-helpers": "~0.0.2",
```
rwjblue: Can we version this so we allow ^ dep? (I think it would require a 0.1.0 at least). That way we have a way to hot fix without requiring both libs to be updated.

pzuraq: Sure thing, I can release `0.1.0` of compatibility helpers

PR #157, opened 2017-08-20
@sort regression when used with property for sort args

NullVoxPopuli:
Related Issue: https://github.com/ember-decorators/ember-decorators/issues/158

This used to work:
```js
  sortBy: 'registeredAt:asc',
  @sort('model.registrations', 'sortBy') registrations: null,
```

but now it gives the error 
 `Assertion Failed: Cannot use 'sort' on field 'registrations' without a callback`

the only way to solve this is to use a full function for the comparator:

```js
  @sort('model.registrations', function(a, b) {
    const ra = a.get('registeredAt');
    const rb = b.get('registeredAt');

    return (ra > rb && 1) || (ra < rb && -1) || 0;
  }) registrations;
```

This PR restores the previous behavior

### DIFF
--- a/addon/utils/decorator-macros.js
+++ b/addon/utils/decorator-macros.js
@@ -1,3 +1,5 @@
+import { IS_EMBER_2 } from 'ember-compatibility-helpers';
+
 import { decoratorWithParams } from './decorator-wrappers';
 import extractValue from './extract-value';
 
@@ -56,7 +58,7 @@ export function decoratedPropertyWithEitherCallbackOrProperty(fn) {
       return fn(...params);
     }
 
-    if (params.length > 1 && lastParamType === 'string') {
+    if (IS_EMBER_2 && params.length > 1 && lastParamType === 'string') {
       assert(`Cannot use '${lastParam}' on field '${key}' because it does not exist on the target`, target[lastParam]);
       return fn(...params);
     }
--- a/addon/utils/decorator-macros.js
+++ b/addon/utils/decorator-macros.js
@@ -50,12 +50,13 @@ export function decoratedPropertyWithEitherCallbackOrProperty(fn) {
     assert(`Cannot use '${fn.name}' on field '${key}' without parameters`, params.length !== 0);
 
     const lastParam = params[params.length - 1]
+    const lastParamType = typeof lastParam;
 
-    if (typeof lastParam === 'function') {
+    if (lastParamType === 'function') {
       return fn(...params);
     }
 
-    if (params.length > 1 && typeof lastParam === 'string') {
+    if (params.length > 1 && lastParamType === 'string') {
       assert(`Cannot use '${lastParam}' on field '${key}' because it does not exist on the target`, target[lastParam]);
       return fn(...params);
     }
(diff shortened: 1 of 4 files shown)


NullVoxPopuli:
ember 1.11 hates me

NullVoxPopuli:
I guess, ember-decorators can drop support for ember 1.x, yeah? :-D

buschtoens:
Probably cause by (my :sweat_smile: ) PR https://github.com/ember-decorators/ember-decorators/pull/139. If so, sorry. I'll look into it.

addon/utils/decorator-macros.js
```diff
@@ -38,8 +38,12 @@ export function decoratedPropertyWithOptionalCallback(fn) {
       return fn(...params);
     }
 
+    if (typeof params[params.length - 1] === 'string' && target[params[params.length - 1]]) {
```
buschtoens: Shouldn't that be `get(target, params[params.length - 1])`?

NullVoxPopuli: I'm just checking existence, and don't care about the value?

NullVoxPopuli:
@buschtoens updated!

addon/utils/decorator-macros.js
```diff
@@ -44,3 +44,25 @@ export function decoratedPropertyWithOptionalCallback(fn) {
...
+  return decoratorWithParams(function(target, key, desc, params) {
+    assert(`Cannot use '${fn.name}' on field '${key}' without parameters`, params.length !== 0);
+
+    const lastParam = params[params.length - 1]
+
+    if (typeof lastParam === 'function') {
+      return fn(...params);
+    }
+
+    if (params.length > 1 && typeof lastParam === 'string') {
```
rwjblue: Seems like we are calling `typeof lastParam` twice, would you mind extracting to a separate variable to track (e.g. `let lastParamType = typeof lastParam`)

NullVoxPopuli: of course!

NullVoxPopuli:
@rwjblue updated!

NullVoxPopuli:
is 1.x support getting dropped? :-D

kellyselden:
@NullVoxPopuli according to the readme this supports ember 2.4+. I wonder which is right?

NullVoxPopuli:
So, if the readme is the most up to date, I guess that means the .travis.yml just needs updating? should that be a part of this PR as well?

rwjblue:
We want to keep 1.11 support if at all possible.  If there is fundamental functionality that doesn't exist in Ember@1.11 that this PR requires, we can easily guard the test and only run it when using 1.13+ but we should look into the failures to confirm the root cause before we exclude the test in 1.11...

rwjblue:
I created https://github.com/ember-decorators/ember-decorators/issues/159 to track getting the README updated.

NullVoxPopuli:
I looked at the computed property sort macro docs the other day for 1.11, and it's identical to the 2.14 docs.. so I'm not sure what's up. :-\

NullVoxPopuli:
@rwjblue I added an `IS_EMBER_2` check to the new condition on the decorator-macros

NullVoxPopuli:
@rwjblue, @kellyselden, @buschtoens   tests pass!

NullVoxPopuli:
@pzuraq thanks!

PR #160, opened 2017-08-23
Follow up to sort regression - es6 support

NullVoxPopuli:
See: https://github.com/ember-decorators/ember-decorators/issues/158#issuecomment-324219503

/cc @buschtoens


NullVoxPopuli:
removed commented out code

PR #165, opened 2017-09-13
Fix markdown in comments

KnownSubset:
Was missing closing ticks.


pzuraq:
Awesome, thanks for the patch! Will deploy the new docs soon

PR #166, opened 2017-09-29
feature(component): Adds component decorators

pzuraq:
Adds decorators for tagName, classNames, classNameBindings, and attributeBindings. These fields, for a variety of reasons, have special behavior when compared to other class fields such as merging, concatenating, or simply having to be applied to the prototype/before init. These decorators provide easy to use, declarative alternatives which should tide us over until we are able to drop them altogether (with Glimmer components).

* `@attribute`: Declares a field an `attributeBinding` and adds it to the array
* `@className`: Declares a field a `classNameBinding` and adds it to the array
* `@classNames`: Adds the class names to the component
* `@tagName`: Specifies the tag of the component

### DIFF
--- a/addon/component/index.js
+++ b/addon/component/index.js
@@ -1,0 +1,114 @@
+import { assert } from '@ember/debug';
+
+import collapseProto from '../utils/collapse-proto';
+import { decoratedConcatenatedProperty } from '../utils/decorator-macros';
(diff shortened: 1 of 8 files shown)


pzuraq:
@rwjblue this is ready to review

PR #167, opened 2017-09-30
feature(legacy): Transform `constructor` into `init` for legacy support

pzuraq:
After dealing with a lot of the edge cases caused by the double-extends injections messing up `constructor`, we're finding that it introduces a lot of overhead and makes ES classes difficult to use. For an example, check out #166, where tests are failing due to the difference in behavior between versions.

This PR introduces a new babel transform for legacy versions of Ember. It renames the `constructor` functions to `init`, allowing class fields and other constructor based features to work as intended. A default constructor that calls `init` is put in place as well to make sure non-Ember Object classes aren't broken.

I think this transform could a separate addon too, open to pulling it out and letting users decide whether or not they want this. I think for writing tests it's going to be invaluable, the amount of branching logic depending on versions will be huge otherwise.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,25 @@
+// launch.json
+{
+  "version": "0.2.0",
+  "configurations": [
+
+  {
+    "type": "node",
+    "request": "launch",
+    "name": "Ember Build",
+    "program": "${workspaceRoot}/node_modules/ember-cli/bin/ember",
+    "args": [
+      "build"
+    ]
+  },
+  {
+    "type": "node",
+    "request": "launch",
+    "name": "Ember Serve",
+    "program": "${workspaceRoot}/node_modules/ember-cli/bin/ember",
+    "args": [
+      "serve"
+    ]
+  }
+  ]
+}
--- a/index.js
+++ b/index.js
@@ -57,11 +57,13 @@ module.exports = {
     let disableTransforms = parentOptions.emberDecorators && parentOptions.emberDecorators.disableTransforms;
 
     if (!this._registeredWithBabel && !disableTransforms) {
-      let checker = new VersionChecker(this.parent).for('ember-cli-babel', 'npm');
+      let emberChecker = new VersionChecker(this.app).forEmber();
+      let babelChecker = new VersionChecker(this.parent).for('ember-cli-babel', 'npm');
 
-      if (checker.satisfies('^6.0.0-beta.1')) {
+      if (babelChecker.satisfies('^6.0.0-beta.1')) {
         let TransformDecoratorsLegacy = requireTransform('babel-plugin-transform-decorators-legacy');
         let TransformClassProperties = requireTransform('babel-plugin-transform-class-properties');
+        let EmberLegacyClassConstructor = requireTransform('babel-plugin-ember-legacy-class-constructor');
 
         // Create babel options if they do not exist
         parentOptions.babel = parentOptions.babel || {};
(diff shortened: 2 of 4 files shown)


rwjblue:
Hmm. I’m unsure about this. We should focus on the underlying issues that you are seeing, and make sure that they definitely require this change. My concern with re-writing constructor to init is that we are painting ourselves in a corner (e.g. in the long run we definitely don’t want to do this, and doing it now makes some things work that shouldn’t)...

pzuraq:
It's definitely an extreme solution so I understand the hesitance. The main issue as I see it is class fields simply do not work pre-2.13. This means a few things:

1. As decorator developers, we must develop two very different codepaths for the same features, one which relies on users being able to use class fields as intended (field value on the instance) and one where they are purposefully hacking around the issue (field value on the prototype). This has non-trivial implications, including:
    1. We have to think through entirely different edge-cases in usage and the way the decorators work with each other and base class
    2. We need to think through an upgrade path for each decorator, otherwise users will end up painted into a corner with the old behavior

2. Users must know where fields should live, and there's nothing really intuitive about this. The `@proto` decorator works, but at times it is unneeded (`@service`, `@controller`, computed property macros) and at other times it is needed. You end up with something like:

```js
@tagName('input')
export default class InputComponent extend Component {
  @attribute @proto type = 'text';
  @className @proto isFocused = false;

  @service focusService;

  @alias('textValue') value;

  @computed
  get foo() {
    //...
  }
}
```

which brings me to point 3

3. It requires the user to learn a _lot_ about the class model to understand which things make sense on the prototype, and which don't, and why. It also affects app code, since the difference in location of fields affects how you interact with the fields (think of the common shared array vs. instance issue in Ember - there are all kinds of tiny little issues like this which pop up), so it's not just at class definition time but whenever you interact with the class at all.

For us these issues add up to making classes very difficult to teach, and an immediate source of tech debt that we know we'll be getting rid of, so using them as is is not what we want.

I'm open to figuring out an alternative for getting class fields to work. We could target the fields with a transform directly, but we'd still need to add something to the constructor or we'll break non-Ember classes. We can try to monkeypatch `Ember.Object.extend`, but I've thought through that before and I don't see a way we can fix everything without completely reworking the internals of CoreObject (not 100% sure though, definitely open to trying). 

If we can't find another way to fix class fields we'll likely end up using this transform, so we can also pull it out into a separate addon like I mentioned and test it out for a while to make sure it's not a total footgun 😛

pzuraq:
Also, FWIW I did think long and hard about the implications of running code during `init` vs the actual constructor. I think they are _close enough_ that there shouldn't be many differences if any at all. They execute in the same context, in a very similar point in class instantiation. 

The two biggest differences are: 

1. the ES Class constructor will now run _before_ any non-ES Class object's `init` function, whereas in modern Ember it would necessarily run once `init` is complete. I could see this having some issues as people upgrade, but most can be mitigated by telling users to always run `super` immediately in the constructor, ensuring that run order would always be the same:

```js
class FooComponent extends Component {
  constructor() {
    super(...arguments);

    // Now do stuff
  }
}
```

2. `finishChains` gets called _after_ `init` but _before_ the constructor would normally run. I don't think this would have massive implications, but if it does we could get around it by adding a function that runs on the `init` _event_, which happens at the very end of the class constructor after everything else.

pzuraq:
Another maybe-less-extreme approach I'm thinking of at the moment would be to transpile classFields to _both_ `init` and `constructor`. Little bit of code duplication, but then there wouldn't have to be any logic added to call `init` from the constructor, and `init` would continue to work as expected. The gotcha there would be users on non-Ember classes would have to know that `init` is off limits, but I think that's as much of an edge case as the other solutions have.

PR #168, opened 2017-10-08
feat(validations): Adds attribute meta info for @attribute and @className

pzuraq:
This PR marks class fields as `attributes` in the validation meta object provided by `@ember-decorators/utils`.

## Why is this needed?

`ember-argument-decorators` by default provides validations on the arguments that are passed into a component. For instance, if you have a component like this:

```js
export default class ExampleComponent extends Component {

}
```

and try to use it like this:

```hbs
{{example-component foo="bar"}}
```

it will throw an error and tell you that the `foo` argument has not been defined. You must declare it as an argument like this:

```js
export default class ExampleComponent extends Component {
  @argument foo;
}
```

`@attribute` and `@className` can be applied to class fields as well, and currently the way they work clashes with the `@argument` decorator (they take the value provided by the field and assign it to the prototype, whereas the `@argument` decorator takes the value and assigns it in the constructor IFF a value has not already been assigned). Because of the difference in how they work, the two decorators cannot be made to work together. It's also undesirable from a DX standpoint - arguments and attributes have a clear distinction in Glimmer, so keeping that distinction in Ember is probably a good idea, even if they end up working similarly.

By adding information to the meta object that `ember-argument-decorators` uses, `ember-argument-decorators` can recognize that these fields exist and whitelist them, treating them similarly to arguments.

The meta information is stripped in production builds, so there won't be any perf implications for users who do not use `ember-argument-decorators`.

## Alternatives

`ember-argument-decorators` can re-export versions of `@attribute` and `@className` which are wrapped to store this meta information. This would likely be confusing for developers, but it can be addressed with thorough documentation.

### DIFF
--- a/addon/utils/decorator-macros.js
+++ b/addon/utils/decorator-macros.js
@@ -1,4 +1,6 @@
 import { IS_EMBER_2 } from 'ember-compatibility-helpers';
+import { validationsForKey } from '@ember-decorators/utils/debug';
+
 
 import { decoratorWithParams } from './decorator-wrappers';
 import extractValue from './extract-value';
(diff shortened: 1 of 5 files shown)


pzuraq:
This actually isn't necessary, we can use the `attributeBindings` and `classNameBindings` properties on the prototype to do the whitelisting. Going to close this.

PR #170, opened 2017-10-10
deprecate(computed): Deprecate directly using @computed on functions

pzuraq:
Adds a deprecation for macro-computeds and using `@computed` directly on functions. When `macro-helper-decorators` has been pulled out the deprecation can be updated to point toward it.

### DIFF
--- a/addon/object/index.js
+++ b/addon/object/index.js
@@ -72,64 +72,13 @@ export const action = decorator(function(target, key, desc) {
 });
 
 /**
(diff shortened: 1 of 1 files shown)


pzuraq:
Updated the docs, I removed the old syntax and replaced it entirely with the new syntax, but I could also keep the old syntax and mark it as deprecated in the docs.

rwjblue:
Nah, let’s just document the new go forward way (like you have here).

PR #176, opened 2017-11-07
[BUGFIX] Missing curly brackets in objects computed documentation

spieker:
The import statement in the example of `computed` where missing the curly brackets. Copy paste of the code results in an error.


pzuraq:
Awesome, thanks for the fix!

Issue #99, opened 2017-06-18
Feature Request: Allow composition of computed decorators


mwpastore:
For feature parity with ember-macro-helpers and ember-awesome-macros.

This works:

```javascript
import EmberObject from 'ember-object';

import or from 'ember-awesome-macros/or';
import computed from 'ember-macro-helpers/computed';

export default EmberObject.extend({
  foo: computed(or('bar', 'qux'), function(barOrQux) {
  })
});
```

This, sadly, does not:

```javascript
import EmberObject from 'ember-object';

import computed, { or } from 'ember-computed-decorators';

export default EmberObject.extend({
  @computed @or('bar', 'qux') // or @computed(or(..))
  foo(barOrQux) {
  }
});
```

This works (with kellyselden/ember-awesome-macros#348):

```javascript
import EmberObject from 'ember-object';

import mapBy from 'ember-awesome-macros/array/map-by';
import sum from 'ember-awesome-macros/sum';
import raw from 'ember-macro-helpers/raw';

export default EmberObject.extend({
  items: [],
  
  totalQuantity: sum(mapBy('items', raw('quantity')))
});
```

This, sadly, does not:

```javascript
import EmberObject from 'ember-object';

import { mapBy, sum } from 'ember-computed-decorators';

export default EmberObject.extend({
  items: [],

  @sum @mapBy('items', 'quantity') // or @sum(mapBy(..))
  totalQuantity
});
```

rwjblue:
Thank you for opening the issue with detailed examples! I _think_ that we can make this work...

pzuraq:
Reopened on the auto-computed decorator repo: https://github.com/ember-decorators/auto-computed/issues/4

PR #177, opened 2017-11-08
:sparkles: Add ability to nest computed properties within a decorator

KnownSubset:
**Did you confirm this fix/feature is something that is needed?** Yep, #99 

**Did you write automated tests?** Yes, I added some of the obvious test cases for the new nested functionality.  If you have ideas on future tests to write, I'll be happy to implement them.

**Did you add documentation for the changes you made?** No, that would be a separate PR

**In lieu of a style guide, did you match your code style to the rest of the project?** Yep

### DIFF
--- a/addon/utils/decorator-wrappers.js
+++ b/addon/utils/decorator-wrappers.js
@@ -24,8 +24,17 @@ export function decoratorWithParams(fn) {
     if (isDescriptor(params[params.length - 1])) {
       return handleDescriptor(...arguments, fn);
     } else {
-      return function(/* target, key, desc */) {
-        return handleDescriptor(...arguments, fn, params);
+      return function(target, key, desc) {
+        const hasDecorator = (typeof params[0]) === "function";
+        const decorators = hasDecorator ? params.filter(param => (typeof param) === "function") : [];
+        const nonDecorators = hasDecorator ? params.filter(param => (typeof param) !== "function") : params;
+        const newProperties = decorators.map((func, index) => {
+          const prop = `_private_${key}_${index}`;
+          Object.defineProperty(target, prop, func(target, prop, desc));
+          return prop;
+        });
+        const properties = [...newProperties, ...nonDecorators];
+        return handleDescriptor(...arguments, fn, properties);
       };
     }
   };
--- a/tests/unit/macro-test.js
+++ b/tests/unit/macro-test.js
@@ -122,11 +122,11 @@ test('bool', function(assert) {
     @bool('null') isNullRad: null
   }).create();
 
-  assert.equal(obj.get('isOneAwesome'),true);
-  assert.equal(obj.get('isZeroSad'),false);
-  assert.equal(obj.get('isStringCool'),true);
-  assert.equal(obj.get('isUndefinedAlright'),false);
-  assert.equal(obj.get('isNullRad'),false);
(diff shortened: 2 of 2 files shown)


pzuraq:
First off, thank you for the awesome PR 😄 always happy 

We definitely do want to support this functionality in some way in the future, but unfortunately things are a bit in flux right now, so it may not be the best time to land this. Using decorators directly

addon/utils/decorator-wrappers.js
```diff
@@ -24,8 +24,17 @@ export function decoratorWithParams(fn) {
...
     } else {
-      return function(/* target, key, desc */) {
-        return handleDescriptor(...arguments, fn, params);
+      return function(target, key, desc) {
+        const hasDecorator = (typeof params[0]) === "function";
+        const decorators = hasDecorator ? params.filter(param => (typeof param) === "function") : [];
+        const nonDecorators = hasDecorator ? params.filter(param => (typeof param) !== "function") : params;
+        const newProperties = decorators.map((func, index) => {
+          const prop = `_private_${key}_${index}`;
+          Object.defineProperty(target, prop, func(target, prop, desc));
```
pzuraq: We probably don't want to modify the prototype like this, since that will be a lasting change. `ember-macro-helpers` already handles a lot of the work for composing computeds, so I think we can probably just 'unwrap' the computeds and pass them directly to the macro helpers version of `@computed`.

snewcomer:
@KnownSubset I really like this PR!

Here is the first commit to [auto-computed](https://github.com/ember-decorators/auto-computed/pull/1
).

addon/utils/decorator-wrappers.js
kellyselden: Correct. `ember-macro-helpers` keeps a proxy object around to define properties on to avoid this situation.

KnownSubset:
@pzuraq I need a bit of guidance. I started to follow your advice to see if we could try to destruct the `ComputedProperty` and pass it to `@computed`.  I quickly started to produce some hacky code to handle the edge cases with `InjectedProperty` before even getting the nested properties working again.

Now that the changes have been made in ember-decorators/ember-decorators@817dec0bfea7b0c4529702340e5e30e3c704794d and the utils have been extracted there does not seem to be much of a point in adding the functionality within this repo.  If we are to add the functionality to `auto-computed` is it encouraged to follow the format of `ember-awesome-macros` where we recreate the functions available from `ember-decorators` like `map-by` and `readOnly`?  If so, would we want them to be able to be defineable only in ES6 get/set format?

pzuraq:
@KnownSubset Sorry I haven't had time to respond to this, between the holidays and getting back I've been really busy. 

I would say no, unless the code in the decorators themselves needs to be substantially different. We could import and re-export the macros from `ember-awesome-macros` for convenience, but other than that we should try to keep it as contained as possible. That said, I have a feeling the reason @kellyselden did that in the first place is because some magic was needed to get those functions to work anyways, so it may make sense.

Macros are currently only applicable to class fields, so I don't think `get/set` syntax is going to be an issue here. `@autoComputed` only applies to functions, which makes sense.

pzuraq:
This functionality has been pulled out into `@ember-decorators/auto-computed`. If you'd like to resubmit it there, I'd be happy to try to help you get it merged!

PR #178, opened 2017-11-08
Update README for new @computed syntax

paulcwatts:
Missed a spot when updating the docs in https://github.com/ember-decorators/ember-decorators/pull/170


rwjblue:
Thank you!

PR #181, opened 2017-11-16
enhancement(attribute): Allow bound attribute name to be specified

### LINKED ISSUE #180: Allow mapped attribute bindings.
In some cases, attributes that need to be bound will include a `-` (or other invalid JS identifier characters). Ember's internal `attributeBindings` syntax supports mapping the local property name to alternate attribute names already, but the `@attribute` decorator does not expose a way to utilize it.

I believe the changes needed would be to add an extra `params` argument to 

https://github.com/ember-decorators/ember-decorators/blob/68362ab0ae3c7f2b7351ec31eac0f23a5bc38077/addon/component/index.js#L30

Then modifying this line

https://github.com/ember-decorators/ember-decorators/blob/68362ab0ae3c7f2b7351ec31eac0f23a5bc38077/addon/component/index.js#L38

To roughly do:

```js
  let binding = params[0] ? `${key}:${params[0]}` : key;
  target.attributeBindings.push(binding);
```

This should enable usage such as:

```js
import Component from '@ember/component';
import { attribute } from 'ember-decorators/component';
import { computed } from 'ember-decorators/object';

export default class AttributeDemoComponent extends Component {
  @attribute('data-foo') foo = 'lol';
}
```

---

Rough outline of the work required:

* Update implementation to support mapping
* Add tests confirming functionality is working properly
* Update inline API docs to have an example of utilizing the mapping

kylemarksmith:
Closes #180

### DIFF
--- a/addon/component/index.js
+++ b/addon/component/index.js
@@ -2,7 +2,7 @@ import { assert } from '@ember/debug';
 
 import collapseProto from '../utils/collapse-proto';
 import extractValue from '../utils/extract-value';
-import { decorator, decoratorWithParams } from '../utils/decorator-wrappers';
+import { decoratorWithParams } from '../utils/decorator-wrappers';
 
 /**
  * Decorator which indicates that the field or computed should be bound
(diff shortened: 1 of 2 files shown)


pzuraq:
🎉  thanks for the PR!

pzuraq:
Released as v1.3.2

Issue #145, opened 2017-07-26
Extract "decorator utils" to separate addon.


rwjblue:
I'm unsure what the name should be, but we should extract some of our building block utilities into a separate addon so that we can share the workload and allow other addons to easily create their own decorators.

The use cases that I am specifically thinking of are:

* @kellyselden's `ember-macro-helpers` (to provide the auto-get `@computed` version once it is deprecated/removed from here #146).
* @dfreeman's `ember-css-modules` based on https://github.com/rwjblue/ember-decorators/issues/133#issuecomment-316711907
* `ember-data` should provide its own decorators for `@attr` / `@hasMany` / etc.

rwjblue:
One question I have is: should we move this repo to an org so that both ember-decorators and the new utils addon live in the same org?

rwjblue:
The use cases that I was specifically thinking of for this issue are:

* @kellyselden's `ember-macro-helpers` (to provide the auto-get `@computed` version once it is deprecated/removed from here #146).
* @dfreeman's `ember-css-modules` based on https://github.com/rwjblue/ember-decorators/issues/133#issuecomment-316711907
* `ember-data` should provide its own decorators for `@attr` / `@hasMany` / etc.

pzuraq:
I think `ember-decorator-macros` would work, that's essentially what they are. I like the idea of it being moved to an org, though it might be overkill to have an org just for the decorators. Would it be possible to use an existing one?

rwjblue:
Not sure.  We have to be somewhat careful to not infer these addons are "official" while decorators are still in Stage 2, and moving the addon(s) to `emberjs` or `ember-cli` "feels" a bit to official to me...

pzuraq:
yeah, fair enough. I'm down to create a new org, or we could move them to `html-next` possibly

buschtoens:
I'm about to implement the aforementioned `ember-css-modules` decorators and would like to use these utils. I'm happy to do the manual labor of extracting these into their own addon, if you can't find the time. :smile: 

I suggest the name `@ember-decorators/utils` or do you have something better in mind?

pzuraq:
Yep, that would be great. We actually already have the repo setup and published an earlier version: https://github.com/ember-decorators/utils

I was planning on using it to share metadata for `@ember-decorators/argument` but it turned out we didn't need to. Feel free to get rid of the -debug folder and stripping, etc.

[...]

PR #182, opened 2017-11-17
refactor: extract @ember-decorators/babel-transforms

### LINKED ISSUE #185: Should display a deprecation notice if the parent does not have `@ember-decorators/babel-transforms` installed
@buschtoens has been extracting the decorators transforms into a separate [addon here](https://github.com/ember-decorators/babel-transforms). For v2 this will be the standard way of adding transforms, and will be included in the default blueprint along with this addon. 

We should add a deprecation notice IFF the parent has not added @ember-decorators/babel-transforms to their project and they are still relying on the transforms in this repository.

buschtoens:
Addressing one half of #145.

I think `@ember-decorators/babel-transforms` is ready for a first release.

You can test drive this PR by installing my `extract-babel-transforms-demo` branch, since it is just the same as this PR, but with a direct GitHub dependency, since `babel-transforms` isn't released yet.

```
ember install ember-decorators@buschtoens/ember-decorators#extract-babel-transforms-demo
```

Supercedes https://github.com/ember-decorators/ember-decorators/pull/187, therefore closes #185.


buschtoens:
The `yarn.lock` has to be re-generated, when `@ember-decorators/babel-transforms` is released. Preferably before merging. :smile:

pzuraq:
So one thing to note about this, we won't be able to merge it until we're ready to release `v2` as it is currently. It's a breaking change since current users will expect ember-decorators to be adding these transforms to their apps.

I wonder if we can just import the `index.js` of `@ember-decorators/babel-transforms` and re-export it with a changed name..

buschtoens:
That's a pretty good idea. 😀
I'll test that, when I'm back at a computer. :)

package.json
```diff
@@ -21,14 +21,12 @@
...
   "dependencies": {
-    "babel-plugin-transform-class-properties": "^6.24.1",
-    "babel-plugin-transform-decorators-legacy": "^1.3.4",
     "ember-cli-babel": "^6.0.0",
-    "ember-cli-version-checker": "^2.0.0",
     "ember-compatibility-helpers": "^0.1.0",
     "ember-macro-helpers": "^0.16.0"
   },
   "devDependencies": {
+    "@ember-decorators/babel-transforms": "^0.0.0",
```
ro0gr: I think this should be a dependency not a devDependency. This way it will be auto installed into a consumer app

buschtoens: Unfortunately, not exactly. If we list it as a `dependency`, this only means that `babel-transforms` will handle transpilation for `ember-decorators`, but not the consumer app / addon.

buschtoens: I go into detail about this in the README of `babel-transforms`. 😊

buschtoens:
@pzuraq  Okay, so unfortunately we cannot leverage prototypes for "extending" `@ember-decorators/babel-transforms`, since ember-cli does not walk the prototype chain when creating the addon instance.

I'm now using `Object.assign`, which works just as well. We have to remember not to overwrite `included(parent)` (or possibly other hooks in the future) though. Shall I add a comment to remind us?

Other than that, I think we're good to go. In https://github.com/ember-decorators/babel-transforms/commit/bdac90abf824bbe1fdbf1d76068f897f4cc875d2 I made sure that we're using the full package name, so if you don't see anything else, you can release the first version of babel-transforms. :slightly_smiling_face:

package.json
pzuraq: I do think @ro0gr is correct though, since we are importing `@ember-decorators/babel-transforms` in `index.js` we are using it now, just not for transforms

buschtoens: Oh, absolutely correct, now that we actually import it. :sweat_smile:

pzuraq:
Looking good, the one last thing I noticed is here:

https://github.com/ember-decorators/babel-transforms/blob/master/index.js#L44-L47

It looks like this changes the options API, and this PR would end up changing them as well, which would be breaking. I agree with the new API in principle, but I think we'll need to shim it for the moment.

I think we can call the `included` function of the transforms plugin directly to make this happen:

```js
const BabelTransforms = require('@ember-decorators/babel-transforms');

module.exports = {
  name: 'ember-decorators',

  included(parent) {
    if (parent.options && parent.options.emberDecorators && parent.options.emberDecorators.disableTransforms) {
      parent.options['@ember-decorators/babel-transforms'] = parent.options['@ember-decorators/babel-transforms'] || {};
      parent.options['@ember-decorators/babel-transforms'].disable = true;
    }

    BabelTransforms.included.apply(this, ...arguments);
  }
}
```

Something like this should work, I think in this case it's alright to modify parent options like that. 

We should also likely add a deprecation warning, ideally it should display a message IFF the parent does _not_ have `@ember-decorators/babel-transforms` installed. I think there may be a way to check the addon list of the parent to see what addons are included, but I'm not sure. We can do this in a followup as well, I'll open an issue for it.

pzuraq:
Also `@ember-decorators/babel-transforms` looks good, I'll go ahead and publish `v0.1.0` today!

buschtoens:
After https://github.com/ember-decorators/babel-transforms/pull/5 is merged and released I can update the `package.json` and `yarn.lock` once more and we should be done with this PR :smile:

buschtoens:
Okay, I'm happy with this PR now. Ready for another review. :smile: 

> After https://github.com/ember-decorators/babel-transforms/pull/5 is merged and released I can update the `package.json` and `yarn.lock` once more and we should be done with this PR :smile: 

Still TBD. :point_up:

pzuraq:
@buschtoens the changes in `babel-transforms` have been released, when you get chance lets update this PR and get it merged 😄

buschtoens:
All done :checkered_flag:

buschtoens:
A pleasure, as always. :smile: :heart:

PR #183, opened 2017-11-20
refactor: extract utils into `@ember-decorators/utils`

buschtoens:
Adressing the other half of #145.

https://github.com/ember-decorators/utils/pull/3

`yarn.lock` needs to be updated.


addon/utils/decorator-macros.js
```diff
@@ -1,70 +0,0 @@
-import { IS_EMBER_2 } from 'ember-compatibility-helpers';
-
-import { decoratorWithParams } from './decorator-wrappers';
-import extractValue from './extract-value';
-
-import { assert } from '@ember/debug';
-
-export function decoratorWithKeyReflection(fn) {
```
pzuraq: I will say that I'm not 💯  on this macro system, it's been very finicky and the fact that we have so many different macros makes me feel like there's gotta be a better abstraction/system. That said, we can save it for the rewrite for v2

pzuraq:
This looks good. Like I said I do think we're going to need to rewrite a lot of these, but it definitely cleans it up to have them be in a separate repo 😄 

If you think `@ember-decorators/utils` is ready to go I can publish v0.2.0

buschtoens:
Nothing that https://github.com/ember-decorators/utils/pull/3 doesn't do has come to mind yet. I'd say go ahead and merge. :+1:

PR #184, opened 2017-11-23
Remove bower install instruction

serchavalos:
I think bower is no longer used in this addon. If so, please merge


rwjblue:
Thank you!

PR #187, opened 2017-12-05
Warning if parent has not installed @ember-decorators/babel-transforms

### LINKED ISSUE #185: Should display a deprecation notice if the parent does not have `@ember-decorators/babel-transforms` installed

snewcomer:
close #185

### DIFF
--- a/index.js
+++ b/index.js
@@ -77,6 +77,13 @@ module.exports = {
         if (!hasPlugin('transform-class-properties')) {
           plugins.push(TransformClassProperties);
         }
+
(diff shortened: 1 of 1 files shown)


pzuraq:
Superceded by #182

PR #188, opened 2017-12-05
Ember Data ES6 macro test

### LINKED ISSUE #98: Ember Data decorators do not appear to work with ES6 classes

snewcomer:
Maybe I'm missing something related to the below issue or just magically works (:man_shrugging:), but here is a passing test.

Fix #98

### DIFF
--- a/tests/unit/ds-macro-test.js
+++ b/tests/unit/ds-macro-test.js
@@ -1,4 +1,4 @@
-import DS from 'ember-data';
+import Model from 'ember-data/model';
 import {
   attr,
   hasMany,
@@ -10,7 +10,7 @@ import { module, test } from 'qunit';
 module('ember-data macro decorator');
 
 test('DS macro', function(assert) {
-  var Model = DS.Model.extend({
+  let ObjectModel = Model.extend({
     @attr firstName: null,
     @attr({ defaultTo: 'blue' }) lastName: null,
     @attr('number') age: null,
@@ -18,14 +18,14 @@ test('DS macro', function(assert) {
     @belongsTo car: null
   });
 
-  Model.store = {
+  ObjectModel.store = {
     modelFor(typeKey) {
       return typeKey;
     }
   };
 
-  var attributes = [];
-  Model.eachAttribute(function(attr, meta) {
+  let attributes = [];
+  ObjectModel.eachAttribute(function(attr, meta) {
     attributes.push({name: attr, type: meta.type});
   });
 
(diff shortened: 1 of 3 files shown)


buschtoens:
#98 is about using ES6 class syntax instead of the Ember Object Model, i.e. this

```js
import { attr } from 'ember-computed-decorators/ember-data';

export default class Post extends Model {
  @attr('string') title;
}
```

instead of this

```js
import { attr } from 'ember-computed-decorators/ember-data';

export default Model.extend({
  @attr('string') title: null
});
```

Your PR adopts [RFC #176](https://github.com/ember-cli/ember-rfc176-data) style imports. :blush:

snewcomer:
@buschtoens oh I think I added a test at the bottom for the ES6 class test..

buschtoens:
Didn't read that carefully enough. Sorry!

Very curious that it seems to work now. I could _swear_ that it did not a few weeks ago.

pzuraq:
I’d like to set up some tests with full Models that have two sided relatioships, really put it through its paces. I have a feeling that the way we are testing here is not triggering whatever the hugs are with the decorators

tests/unit/ds-macro-test.js
```diff
@@ -35,10 +35,101 @@ test('DS macro', function(assert) {
...
+
+  let user_relationships = [];
+  User.eachRelationship(function(attr, meta) {
+    user_relationships.push({key: attr, type: meta.type});
+  });
+
+  assert.deepEqual(user_relationships, [
+    {key: 'storefronts', type: 'storefront'}
+  ]);
+});
```
snewcomer: Next, I could take a shot at registering models with the registry, similar to [this](https://github.com/emberjs/data/blob/master/tests/helpers/store.js) and then do a store.push...then store.find to ensure the relationships are setup correct. Thoughts?

tests/unit/ds-macro-test.js
```diff
@@ -1,44 +1,236 @@
...
+    });
+  });
+
+  return run(() => {
+    const tag = store.peekRecord('tag', 5);
+    assert.equal(get(tag, 'name'), 'friendly', 'retrieves tag record from store');
+    const user = store.peekRecord('user', 1);
+    assert.equal(get(user, 'name'), 'Tom Dale', 'retrieves user record from store');
+    // assert.equal(get(user, 'tag') instanceof Tag, true, 'the tag property should return a tag');
+    // assert.equal(get(user, 'tag.name'), 'friendly', 'the tag should have name');
```
snewcomer: @buschtoens @pzuraq Ah here we go.  So `parentType` is not defined for the relationship model (ED [line](https://github.com/emberjs/data/blob/0999f8a0a756a3118f1e6a4e110709be72e1ad3e/addon/-private/system/model/model.js#L1323)).  Maybe ya'll already knew this :)  See the test above that shows pushing/finding the relationship with object protos work whereas this test doesn't.

pzuraq: Sweet, thanks for tracking this down! Will help a lot with figuring out a fix. I should be able to get to this sometime after the holidays

pzuraq:
This was superceded by the work to breakout @ember-decorators/data

PR #190, opened 2017-12-14
bind action to target

snewcomer:
When dealing with closure actions, I found a use case where the context of the function was undefined. 

I added a test that shows the context is FooComponent, but I need to write a test where it isn't (or maybe make a dummy application to reproduce)

### DIFF
--- a/addon/object/index.js
+++ b/addon/object/index.js
@@ -18,7 +18,10 @@ import {
   SUPPORTS_NEW_COMPUTED
 } from 'ember-compatibility-helpers';
 
-const { computed: emberComputed } = Ember;
+const { 
+  computed: emberComputed,
+  run: { bind }
+} = Ember;
 
 /**
  * Decorator that turns the target function into an Action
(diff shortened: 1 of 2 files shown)


snewcomer:
This is my fault.  Was not pulling the action off of the actions hash in the template.  :man_facepalming:

snewcomer:
This is my fault.  Was not pulling the action off of the actions hash in the template.  :man_facepalming:

pzuraq:
What do you mean exactly by pulling the action off the actions hash?

snewcomer:
I was doing `{{my-comp showAction=showAction}}` instead of `{{my-comp showAction=(action "showAction")}}`.  So the template found something apparently, which seemed strange to me - usually see (i.e. no `Assertion Failed: Action passed is null or undefined in...` error)

buschtoens:
> So the template found something apparently, which seemed strange to me - usually see (i.e. no Assertion Failed: Action passed is null or undefined in... error)

I assume that is your component?

```js
export default class FooComponent extends Component {
  @action
  showAction() {
    // ...
  }
}
```

The `@action` helper only adds an action of the same name that calls the original method. This means that you now have an action named `showAction` and a method named `showAction`, which is why there's no assertion thrown.

PR #193, opened 2018-01-02
Remove redundant

denzo:


pzuraq:
Thank you for the PR! The failures appear to be unrelated and fixed by #194, can you rebase this when you have a moment so we can get it merged?

PR #196, opened 2018-01-08
bugfix(node): Fixes the node version requirement

pzuraq:
In the upgrade to Ember-CLI 2.17 we accidentally messed up the Node version pin. This PR updates the pin and sets the test suite to run against the minimum version of Node.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     "loader.js": "^4.2.3"
   },
   "engines": {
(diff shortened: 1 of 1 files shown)


pzuraq:
The reason for the Node version being set to v6 is for ESDoc. I'll try to dive into this at some point soon, but for now going to reset it to get the fix out.

Issue #142, opened 2017-07-25
Overwrite default blueprints with ES6 classes version


buschtoens:
https://github.com/rwjblue/ember-decorators/pull/137#discussion_r129353784

> We should consider adding some blueprints that generate the new style and would add the classname. Definitely agree that we should use realistic class names

I would suggest that the name for classes should be `<LastPathSegment><Type>`, in order to avoid absurdly long class names for nested classes. At least for components, since Module Unification will disallow nested components anyway (local lookup components are a different thing).

What about routes, controllers, etc? Should the name also be just `<LastPathSegment><Type>` or should we use the full path?

```js
// app/components/user-table/user-avatar/component.js
// ember g component user-table/user-avatar
import Component from '@ember/component';

export default class UserAvatarComponent extends Component {

}
```

```js
// app/profile/likes/route.js
// ember g route profile/likes
import Route from '@ember/routing/route';

export default class LikesRoute extends Component {

}
```

pzuraq:
I agree regarding components that we should use last path segment. For routes/controllers it's tricky, since there could be duplicate class names introduced (`IndexController/Route`, etc.) which could be confusing. 

To be fair, users can name the classes whatever they want, so if they don't like the default they can change it.

buschtoens:
You can also have nested components that have the same last path segment as other nested components. So you'll run into the same problem. 

But I think that this is an engineering problem, that should be solved by the doc generator, i.e. it should take the directory structure into account.

Looking at this repo, ESDoc appears to do so. YUIDoc uses `@module` tags for hierarchy. I don't know about JSDoc and TypeDoc. But I wanted to give the latter one another try, seeing that they seemed to have fixed some blocking issues. 

Anyways, still:

> To be fair, users can name the classes whatever they want, so if they don't like the default they can change it.

Maybe we can even make the name generation configurable / hookable, or do you think that's a bit over-engineered?

pzuraq:
I think for now it's over-engineered, we should pick one and go with it and see what people prefer in the wild. I'd want to see what @rwjblue thinks, but I'd consider updating blueprints to do a different style in the future would not be a breaking change, so we could switch if users didn't like our pick.

chilicoder:
Any updates here? Is it possible to prepare blueprints to be able to create both versions (one according to some configuration)?

pzuraq:
[...]

PR #209, opened 2018-04-04
142 es6 classes blueprint [components, routes, services]

allthesignals:
Work in progress PR to override the component, routes, and services generator blueprints with one that outputs es6 classes. Addresses #142


blueprints/component/index.js
```diff
@@ -0,0 +1,118 @@
+'use strict';
```
chilicoder: do we really need this files? if we won't create them won't they be taken from `ember-source`?

allthesignals: Cool! Great, yes, you're right. I can remove those. I wasn't sure if/how things get picked up at root level.

allthesignals: @chilicoder hmm I still need the template files it appears, and the component file needs `importTemplate` which is referenced in the template. Otherwise, it's not found. Probably need to find a more sophisticated way to extending these from ember-source...

chilicoder: strange.. will check it later

chilicoder: Yeah. I have the same issue. When I have time will debug `ember generate` command

chilicoder: @allthesignals I found how to solve this. inside of your `component/index.js` write
```
const Blueprint = require('ember-source/blueprints/component');

module.exports = Blueprint;
```

allthesignals: @chilicoder sweet! I'll update. Thank you :) figured it was just the classic AMD stuff

pzuraq:
@allthesignals based on the convo above it looks like you still had changes you wanted to make here?

allthesignals:
@pzuraq nope - @chilicoder merged this branch into one of his PRs :)

PR #211, opened 2018-04-04
142 es6 classes blueprint for ember-data [adapter, serializer, transform, model]

chilicoder:
Addresses #142

### DIFF
--- a/blueprints/model/HELP.md
+++ b/blueprints/model/HELP.md
@@ -1,0 +1,27 @@
+<grey>You may generate models with as many attrs as you would like to pass. The following attribute types are supported:</grey>
+  <yellow><attr-name></yellow>
+  <yellow><attr-name></yellow>:array
+  <yellow><attr-name></yellow>:boolean
+  <yellow><attr-name></yellow>:date
+  <yellow><attr-name></yellow>:object
+  <yellow><attr-name></yellow>:number
+  <yellow><attr-name></yellow>:string
+  <yellow><attr-name></yellow>:your-custom-transform
+  <yellow><attr-name></yellow>:belongs-to:<yellow><model-name></yellow>
+  <yellow><attr-name></yellow>:has-many:<yellow><model-name></yellow>
+
+For instance: <green>\`ember generate model taco filling:belongs-to:protein toppings:has-many:toppings name:string price:number misc\`</green>
+would result in the following model:
+
+```js
+import DS from 'ember-data';
+const { Model, belongsTo, hasMany, attr} = DS;
+
+export default class TacoModel extends Model {
+  filling = belongsTo('protein');
+  toppings = DS.hasMany('topping');
+  name = attr('string');
+  price = attr('number');
+  misc = attr()
+}
+```
--- a/blueprints/model/index.js
+++ b/blueprints/model/index.js
@@ -1,0 +1,78 @@
+/* eslint-env node */
+
+var inflection  = require('inflection');
+var stringUtils = require('ember-cli-string-utils');
+var EOL         = require('os').EOL;
+
+module.exports = {
+  description: 'Generates an ember-data model.',
+
+  anonymousOptions: [
+    'name',
+    'attr:type'
+  ],
+
+  locals: function(options) {
+    var attrs = [];
+    var needs = [];
+    var entityOptions = options.entity.options;
+
+    for (var name in entityOptions) {
+      var type = entityOptions[name] || '';
+      var foreignModel = name;
+      if (type.indexOf(':') > -1) {
+        foreignModel = type.split(':')[1];
+        type = type.split(':')[0];
+      }
+      var dasherizedName = stringUtils.dasherize(name);
+      var camelizedName = stringUtils.camelize(name);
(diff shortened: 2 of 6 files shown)


allthesignals:
@chilicoder I had some trouble with these. 

_Model_
I'm not sure if this is related to class syntax, decorators, or the way the blueprint names the model class, but my model wasn't being picked up by the Ember inspector even when it had data in it. There was some weird behavior. Thoughts?

chilicoder:
I'll have a look

allthesignals:
@chilicoder @pzuraq and I have been chatting about this, it's probably just related to the way Ember inspector picks up things. I'll have to make an issue in the inspector. In my real-world usage of the class syntax though, it works fine otherwise.

chilicoder:
I created sample example
```
~/Repos/ember-decorators git:(142-es6-classes-blueprint-emberdata) ✗ ember g model some-model
installing model
  create addon/models/some-model.js
installing addon-import
  create app/models/some-model.js
```
it somehow thinks that I'm in addon

allthesignals:
Yeah, that's normal behavior because you're in an addon.

chilicoder:
@allthesignals I deleted my comment because I accidentally run the command in the wrong folder

chilicoder:
I also checked. Models work in my sample app. You can find it in TreeView of Inspector but not in Data

chilicoder:
I added tests for blueprints. Currently they check only simple case for model, but I plan to add more test for the cases when the user specifies attributes 

PS. do we really need this sophistication for this addon? I see this feature (additionally providing attributes) as really rarely used.

chilicoder:
Hi, can anybody give me feedback for this PR. What we are missing to merge it?

pzuraq:
Hey @chilicoder, I'm really sorry I haven't gotten around to it, been very busy the past few weeks and my priority in my open source time has been to get Ember Decorators v2 out, and then to fix up all the documentation surrounding it. It's definitely not fair to you, my bad 😞Let's get this merged before the monorepo stuff comes through, along with the other blueprints PRs. I'm reviewing right now.

node-tests/blueprints/model-test.js
```diff
@@ -0,0 +1,29 @@
...
+  setupTestHooks(this);
+
+  it('model foo', function() {
+    let args = ['model', 'foo'];
+
+    // pass any additional command line options in the arguments array
+    return emberNew()
+      .then(() => emberGenerateDestroy(args, (file) => {
+        expect(file('app/models/foo.js')).to.contain('import DS from \'ember-data\';'+EOL +
+          ''+EOL +
```
pzuraq: This feels a little messy, but I can't think of a cleaner way to do it right now..

blueprints/model/index.js
```diff
@@ -0,0 +1,78 @@
...
+      if (/has-many|belongs-to/.test(dasherizedType)) {
+        needs.push("'model:" + dasherizedForeignModelSingular + "'");
+      }
+    }
+
+    var needsDeduplicated = needs.filter(function(need, i) {
+      return needs.indexOf(need) === i;
+    });
+
+    attrs = attrs.join('; ' + EOL + '  ');
```
pzuraq: We should add a trailing semi-colon to the last attribute

node-tests/blueprints/model-test.js
chilicoder: I totally copied approach from `ember-simple-oauth`
https://github.com/simplabs/ember-simple-auth/tree/master/node-tests/blueprints

chilicoder:
About module unification. I took index.js file from `ember-data` repo and patched it to provide stuff in the right format. there is no module-unification blueprints currently in ember-data. After they will add it we can copy approach and add them here.

blueprints/model/index.js
chilicoder: Its totally wrong. I will redo

chilicoder:
@pzuraq I pushed another commit with fixed generators for model with attributes. Now this command `ember generate model taco filling:belongs-to:protein toppings:has-many:toppings name:string price:number misc` will correctly generate the file.

Also, I refactored `index.js` file. Instead of rewriting, I import blueprint`s index and transform `attrs` property. Also, I'm adding `generatorImports` property to import needed generators.

blueprints/model/HELP.md
```diff
@@ -0,0 +1,28 @@
...
+For instance: <green>\`ember generate model taco filling:belongs-to:protein toppings:has-many:toppings name:string price:number misc\`</green>
+would result in the following model:
+
+```js
+import DS from 'ember-data';
+const { Model } = DS;
+import { belongsTo, hasMany, attr } from 'ember-decorators/data';
+
+export default class TacoModel extends Model {
+  @belongsTo('protein') filling(){};
```
pzuraq: So the correct output here would be:

```
@belongsTo('protein') filling;
@hasMany('topping') toppings;
@attr('string') name;
```

They should not be functions, but class properties, and they should each end with a semicolon.

chilicoder:
@pzuraq refactored

chilicoder:
@pzuraq my pleasure!

PR #217, opened 2018-04-23
Update readme for 2.0

kernel-io:
import path has changed in v2


kernel-io:
@pzuraq thanks for the answer, I've updated the README.md to reflect the new 2.0 import path

PR #219, opened 2018-04-29
deleted private true repo

chilicoder:


chilicoder:
looks like travis doesn't run otherwise

pzuraq:
Ah that makes sense, probably why orbit.js doesn’t use Travis

pzuraq:
Looks like Travis is running on #220 despite this, so maybe something else happened with your other PR that's causing it not to run. Can you try squashing the commits and seeing if that helps?

chilicoder:
I just reopened it, travis goes

PR #226, opened 2018-05-09
[docs] Adds Addon Docs

pzuraq:
Adds Addon Docs to document decorators. A new set of guides has been
written, attempting to be the definitive guide on using native classes
in Ember today, and the doc comments have all been updated and reviewed.

screen shot 2018-05-08 at 8 04 44 pm


packages/-ember-decorators/tests/dummy/config/environment.js
```diff
@@ -44,6 +45,8 @@ module.exports = function(environment) {
   }
 
   if (environment === 'production') {
+    // Allow ember-cli-addon-docs to update the rootURL in compiled assets
+    ENV.rootURL = 'ADDON_DOCS_ROOT_URL';
     ENV.locationType = 'hash';
     ENV.rootURL = '/ember-decorators/';
```
dfreeman: Looks like these `rootURL`s are going to conflict

pzuraq: good catch, ty!

PR #228, opened 2018-05-09
Fix data decorators type defs

jamescdavis:
The TypeScript type defs needed some work. In particular:

* The type argument to `@attr` should extend the `TransformRegistry` instead of the `ModelRegistry`.
* The definitions do not allow for options to be specified.

As of https://github.com/DefinitelyTyped/DefinitelyTyped/pull/25282, `RelationshipOptions` is now available on the DS namespace. `AttrOptions` was previously available, but incomplete. `@types/ember-data@2.14.17` includes these updates.

We're using almost these exact definitions in a production app [here](https://github.com/CenterForOpenScience/ember-osf-web/blob/develop/types/%40ember-decorators/data.d.ts) (we've just omitted the no-arg defs because we always supply at least the type). I have tested these exact definitions locally with our app.


jamescdavis:
@chriskrycho ☝️

packages/data/index.d.ts
```diff
@@ -57,15 +60,18 @@ export function hasMany(): PropertyDecorator;
...
  * });
  * ```
  *
  * @function
  * @param {String} [type] - Type of the `hasMany` relationship
+ * @param {Object} [options] - Optional relationship options
  */
 export function hasMany<K extends keyof ModelRegistry>(
-  type: K
+  type: K,
```
chriskrycho: This is why I got the `attr` one wrong originally! I started by building out `hasMany` and `belongsTo` and didn’t flip the mental switch when I switched to `attr`.

jamescdavis: That's what I guessed had happened. It'd probably be better to call the first argument something like `transform` for `attr` and `modelName` for `hasMany`/`belongsTo`, but I left them as `type` for now to be consistent with the doc blocks in https://github.com/ember-decorators/ember-decorators/blob/master/packages/data/addon/index.js

`@types/ember-data` is inconsistent for [hasMany](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/ember-data/index.d.ts#L82) and [belongsTo](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/ember-data/index.d.ts#L70).

jamescdavis: Also, I intentionally omitted the Sync/Async distinction that `@types/ember-data` has because we can't currently affect the attribute's type with decorators anyways.

jamescdavis:
Thanks, @pzuraq !

PR #236, opened 2018-05-18
Add main dependency to example

allthesignals:
The example here should be clear that it adds the main `ember-decorators` addon and not only the other two dependencies.


pzuraq:
Master has been red for awhile now and I haven't been able to dig in and reproduce, everything works locally so unsure what's happening 😕 I'm going to try to dig in soon

pzuraq:
Hey @allthesignals, if you can rebase this PR tests should be green noa!

PR #247, opened 2018-06-20
[feat] add @macro util for creating decorators from macros (#213)

### LINKED ISSUE #213: @macro decorator/function
Currently there is no easily accessible way to make our own macro decorators. We should provide a function which takes existing computed macros (such as those in [ember-awesome-macros](https://github.com/kellyselden/ember-awesome-macros)) and converts them into decorators. The helper will be invokable directly and indirectly by binding to a different variable, e.g.

```js
let foo = macro(bar);

class Foo {
  @foo baz;
  @macro(bar) qux;
}
```

buschtoens:
Implements a utility function for creating decorators from computed property macros. Both styles as discussed in #213 have been implemented:

```js
let foo = macro(bar);

class Foo {
  @foo('a', 'b') baz;
  @macro(bar)('c', 'd') qux;
}
```

```js
let foo = macro(bar, 'a', 'b');

class Foo {
  @foo baz;
  @macro(bar, 'c', 'd') qux;
}
```

Closes #213.

### DIFF
--- a/packages/-ember-decorators/tests/unit/object/macro-test.js
+++ b/packages/-ember-decorators/tests/unit/object/macro-test.js
@@ -1,8 +1,9 @@
-import { get, set } from '@ember/object';
+import { get, set, computed } from '@ember/object';
 import { A as emberA } from '@ember/array';
 import { gte } from 'ember-compatibility-helpers';
 
 import {
+  macro,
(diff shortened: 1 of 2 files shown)


allthesignals:
Awesome!!!

PR #250, opened 2018-06-27
fix: in blueprints don't assume ember-data exists

lennyburdette:
Before this change, running `ember generate --help` with `ember-decorators` installed would fail if `ember-data` is not also installed.


lennyburdette:
Looks like the "ember-default" run just needs to be restarted. I don't know if the beta and canary runs are expected to pass?

pzuraq:
There may have been changes in beta and canary that cause breakage, one way that we make sure we address bugs early is making them required for CI. I’ll have some time to dive into this tomorrow, will see if we can get this merged and a patch released!

rwjblue:
This is the failure on beta/canary, and _definitely_ feels related to the object model changes that landed in beta recently:

```
not ok 24 Chrome 67.0 - typescript | @attribute: decorator works correctly through traditional and ES6 hierarchy
    ---
        actual: >
            1
        expected: >
            2
        stack: >
not ok 32 Chrome 67.0 - typescript | @className: decorator works correctly through traditional and ES6 hierarchy
    ---
        actual: >
            1
        expected: >
            2
        stack: >
not ok 40 Chrome 67.0 - typescript | @classNames: decorator works correctly through traditional and ES6 hierarchy
    ---
        actual: >
            1
        expected: >
            2
        stack: >
not ok 74 Chrome 67.0 - typescript | @action: actions are properly merged through traditional and ES6 prototype hierarchy
    ---
        actual: >
            null
        stack: >
```

pzuraq:
@lennyburdette the upstream bugs should be fixed now, any chance you could rebase this?

PR #258, opened 2018-07-21
Add type definition for the macro() function

lsg-richard:
The tsdoc content is copied verbatim from `computed.js`.

### DIFF
--- a/packages/object/computed.d.ts
+++ b/packages/object/computed.d.ts
@@ -381,7 +381,7 @@ export function lte(dependentKey: string, value: Number): PropertyDecorator;
 
   @param {function} fn - The macro function to create a decorator from
   @param {...any} params - Parameters to be partially applied to the macro fn
-  @return {ComputedProperty}
+  @return {PropertyDecorator}
  */
 export function macro<TResult>(fn: () => TResult): PropertyDecorator;
 export function macro<T1, TResult>(fn: (a: T1) => TResult): (a: T1) => PropertyDecorator;
--- a/packages/object/computed.d.ts
+++ b/packages/object/computed.d.ts
@@ -341,6 +341,70 @@ export function lt(dependentKey: string, value: Number): PropertyDecorator;
  */
 export function lte(dependentKey: string, value: Number): PropertyDecorator;
 /**
+  Creates a new decorator from a computed macro function. For instance, you can
+  use this utility function to create decorators from the macros provided by
+  addons such as [ember-awesome-macros](https://github.com/kellyselden/ember-awesome-macros).
+
+  ```js
+  import { macro } from '@ember-decorators/object/computed';
+  import firstMacro from 'ember-awesome-macros/array/first';
+
+  const first = macro(firstMacro);
+
+  export default class LeaderBoardComponent extends Component {
+    ranking = ['Natalie', 'Emma', 'Thomas'];
+
+    @first('ranking') winner; // => 'Natalie'
+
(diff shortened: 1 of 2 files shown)


lsg-richard:
Have updated to add full type checking when currying arguments using `macro()`.

It currently supports up to 5 arguments on the passed `fn`. AFAIK this can't be done generically for N arguments in TypeScript at present. This may be addressed by https://github.com/Microsoft/TypeScript/issues/5453, but I'm not certain and it's also a way off.

Not 100% sure about the `@return {ComputedProperty}` in the TSDoc content - is this reasonable for `macro()` now?

packages/object/computed.d.ts
```diff
@@ -340,6 +340,70 @@ export function lt(dependentKey: string, value: Number): PropertyDecorator;
...
+  export default class NewsPaperComponent extends Component {
+    title = 'Ember chosen best framework of the year again';
+
+    @newsFlash('title') attentionGrabber; // => 'News Flash: EMBER CHOSEN BEST FRAMEWORK OF THE YEAR AGAIN!'
+  }
+  ```
+
+  @param {function} fn - The macro function to create a decorator from
+  @param {...any} params - Parameters to be partially applied to the macro fn
+  @return {ComputedProperty}
```
buschtoens: You're right, this should be `PropertyDecorator`. Can you change this in the actual source code as well?

packages/object/computed.d.ts
```diff
@@ -340,6 +340,70 @@ export function lt(dependentKey: string, value: Number): PropertyDecorator;
...
+  const newsFlash = macro(titleGeneratorMacro, 'News Flash');
+
+  export default class NewsPaperComponent extends Component {
+    title = 'Ember chosen best framework of the year again';
+
+    @newsFlash('title') attentionGrabber; // => 'News Flash: EMBER CHOSEN BEST FRAMEWORK OF THE YEAR AGAIN!'
+  }
+  ```
+
+  @param {function} fn - The macro function to create a decorator from
```
buschtoens: I _think_ this also needs an `@function` like the other doc blocks.

lsg-richard: Would a `@function` tag be redundant in `computed.d.ts`? I couldn't actually see any mention of this tag in TSDoc, but may have missed it.

Looking at `computed.js`, the `@function` tags were added to exports that weren't self-evidently functions, but `macro()` is defined as a function directly in the source file, so again it may be redundant.

Just not 100% sure of the documentation semantics here.

buschtoens: > Looking at computed.js, the @function tags were added to exports that weren't self-evidently functions,

Yes, that's why I did not add the tag there.

> but macro() is defined as a function directly in the source file, so again it may be redundant.

If TS is smart enough to deduce this, I guess we can skip it here. By that logic, we can probably skip it in all of `computed.d.ts` then. 🤔

lsg-richard: Assuming this is a copy-paste job it's probably easier to just keep it the same in both?

pzuraq: I don't think it matters much, since we aren't using the output of TypeDoc yet. We can consider switching over in the future since it'll be able to infer even more.

pzuraq:
This looks good, I'm not a TS expert but I agree I don't think it's possible to type this well until TS3. 5 arguments is probably good for 90% of use cases, so I think we can use it for now.

PR #259, opened 2018-07-23
@layout decorator for components in /addon

mike-north:
When generating a component in an addon, it's currently required to explicitly specify the layout since it won't automatically be resolved while living in the `/addon` folder

##### (A) ES5 - `ember g component x-foo --in-repo=my-addon`
```js
import Component from '@ember/component';
import layout from '../templates/components/x-foo';

export default Component.extend({
  layout
});
```
As people start to use ES6/TS, they're tempted to incorrectly consider `Component.extend { }` to be equivalent to `class extends Component { }`. This leads to the following bad patterns

##### (B) ES6/TS - common mistake of placing the layout on each instance, instead of the prototype
```ts
import Component from '@ember/component';
import layout from '../templates/components/x-foo';

export default class extends Component {
  layout = layout;
}
// or
export default class extends Component {
  constructor() {
    super(...arguments);
    this.layout = layout;
  }
}
```
To maintain equivalence with *(A)* while using ES6 or TS, the following is required

##### (C) ES6/TS - The best we can do today, to create the equivalent of (A) with classes
```ts
import Component from '@ember/component';
import layout from '../templates/components/x-foo';

export default class Foo extends Component {}

Foo.prototype.layout = layout; // 🤮
```
This change adds a new `@layout` decorator. It accepts exactly one argument, which should be a compiled template (either imported or inline-precompiled).

##### (D) The `@layout` decorator
  ```js
  import template from '../templates/components/x-foo';

  @layout(template)
  export default class TagNameDemoComponent extends Component {}
  ```

  ```js
  import hbs from 'htmlbars-inline-precompile';

  @layout(hbs`<h1>Hello {{ name }}</h1>`)
  export default class TagNameDemoComponent extends Component {
    constructor() {
      super(...arguments);
      this.set('name', 'Tomster');
    }
  }
  ```

### DIFF
--- a/packages/component/addon/index.js
+++ b/packages/component/addon/index.js
@@ -153,3 +153,39 @@ export function tagName(tagName) {
     return klass;
   }
 }
+
+/**
+  Class decorator which specifies the layout for the component. This replaces
+  the `layout` property on components in the traditional Ember object model.
+
+  ```js
+  import template from '../templates/components/x-foo';
+
+  @layout(template)
+  export default class TagNameDemoComponent extends Component {}
+  ```
+
+  ```js
+  import hbs from 'htmlbars-inline-precompile';
+
+  @layout(hbs`<h1>Hello {{ name }}</h1>`)
+  export default class TagNameDemoComponent extends Component {
+    constructor() {
+      super(...arguments);
+      this.set('name', 'Tomster');
+    }
+  }
+  ```
+
+  @param {TemplateFactory} template - The compiled template to be used for the component
+*/
+export function layout(template) {
+  assert(`The @layout decorator must be provided exactly one argument, received: ${template}`, arguments.length === 1);
+  assert(`The @layout decorator must be provided a template, received: ${template}. If you want to compile strings to templates, be sure to use 'htmlbars-inline-precompile'`, typeof template !== 'string');
+  assert(`The @layout decorator must be provided a template, received: ${template}`, (() => typeof template === 'object' && typeof template.indexOf === 'undefined')());
+
+  return function(klass) {
+    klass.prototype.layout = template;
+    return klass;
+  }
+}
--- a/packages/-ember-decorators/tests/unit/component/layout-test.js
+++ b/packages/-ember-decorators/tests/unit/component/layout-test.js
@@ -1,0 +1,81 @@
+import Component from '@ember/component';
+import { layout } from '@ember-decorators/component';
+
+import hbs from 'htmlbars-inline-precompile';
+import { moduleForComponent } from 'ember-qunit';
+import { test } from 'qunit';
+import { find } from 'ember-native-dom-helpers';
+
+moduleForComponent('javascript | @layout', { integration: true });
+
+test('decorator sets layout of component', function(assert) {
+  @layout(hbs`<section class='foo'>Hello, world!</section>`)
+  class FooComponent extends Component {}
+
+  this.register('component:foo-component', FooComponent);
+
+  this.render(hbs`{{foo-component}}`);
+  assert.ok(find('section.foo'));
+});
+
+test('decorator throws an error if given a non-template value', function(assert) {
+  assert.throws(
+    () => {
+      @layout([`{{foo-component}}`])
+      class FooComponent extends Component {}
+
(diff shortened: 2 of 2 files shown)


packages/component/addon/index.js
```diff
@@ -153,3 +154,46 @@ export function tagName(tagName) {
...
+  return function(klass) {
+    klass.prototype.layout = template;
+    return klass;
+  }
+}
+
+/**
+ * Test to see whether a given value looks like a compiled handlebars template
+ */
+function isTemplate(val) {
```
rwjblue: if you move this inline (into the assert) it will be stripped in production, I'm unsure if dead code elimination will strip it as it is now

pzuraq:
We actually did consider this and didn't add the `@layout` decorator in the past, because placing the layout on each instance is not actually a problem from a performance perspective since (in general) we aren't actually creating a new instance of the object:

```js
const arr = [];

class Foo {
  newArr = []; // Creates a new array for each instance
  sameArr = arr; // Assigns the same array to every of the class
}
```

We haven't been historically shooting for parity with the object model in terms of _exact_ behavior (otherwise, fields would be unusable entirely) but instead in terms of actual, expected end behavior, and placing the layout on each instance just happened to work and be less code/work. It is admittedly a bit awkward semantically, given that there are other features which are decorators (`@tagName`, `@classNames`).

Is there any reason to add this decorator other than maintaining the exact same behavior as before? Are there any subtle differences in assigning the value to the prototype vs the instance that we haven't accounted for? Not for or against it at this point, just trying to be clear about the reasoning.

packages/component/addon/index.js
mike-north: This began as a more complex check on the value. Do you think that checking for `type = 'object'` is good enough? We could always wrap a more complete assertion in `runInDebug`

rwjblue: I was thinking you could move the function into the assert directly:

```js
assert('asdfasdfa', (() => {
  // function body here...
})());
```

mike-north:
@pzuraq I do understand that, in the case of layout's the current version of Ember results in *N references to 1 object*. It's super clear that things like `@tagName` save us from allocating N strings.

My case for offering this is as follows:

* confusion between `.extend({})` and `class extends C { }` is an extremely common pain point as developers transition to ES6/TS. Having worked with a fairly large number of them, my guess would be way over 50% of people hitting this problem.

* Although you're not aiming for feature parity, parity is _almost_ possible. This is literally the last thing that developers typically encounter, that should exist on a prototype, and no decorator exists yet. (I'm less worried about less-commonly-customized things like `concatenatedProperties: []` and `mergedProperties: []`). Addon authors are left feeling like they have a 95% solution

* In inline hbs cases, you absolutely will end up with N templates instead of 1, unless there's some special de-duping magic happening behind the scenes
```ts
class {
   layout = hbs`<h1> hello world </h1>`
}
```
I guess we could direct every one to define top-level-module-scope vars for this, but a decorator honestly seems easier
```ts
const layout = hbs`<h1> hello world </h1>`
class {
   layout
}
```

pzuraq:
Definitely agree with the inline case, I ran into that myself recently. I think if we can prevent that footgun this feature is worth it. Hopefully this won't be an issue much longer anyways so let's add this for now and move forward.

pzuraq:
Thanks for the feature! Really appreciate it 😄

gossi:
I brainlessly do run into this over and over again. It is even more annoying when you subclass a component where the layout isn't on the prototype and you do have to "assign" the template from the parent component to the one of your subclassed child component (which you'd except from a `.extend({})` behavior). So this comes in very handy and will eliminate all this pain.

Thanks a lot, I'm looking forward to it.

pzuraq:
@gossi that shouldn't be an issue with subclassing though, because class fields are also inherited:

```js
class Foo {
  foo = 123;
}

class Bar extends Foo {
  bar = 456;
}

let bar = new Bar();
bar.foo; // 123
```

So layouts used as fields should work as is and be inherited. It sounds like something else may be going wrong with your setup

pzuraq:
Released in v2.3.0!

PR #262, opened 2018-07-27
[bugfix] Return promise in default blueprint

pzuraq:


pzuraq:
Hmm, that's some nasty flakiness. Will have to look into it..

knownasilya:
Ready to merge?

PR #266, opened 2018-08-05
additional type for attr

akatov:
it can also accept just the options hash

### DIFF
--- a/packages/data/index.d.ts
+++ b/packages/data/index.d.ts
@@ -23,6 +23,22 @@ export function attr(): PropertyDecorator;
  * import { attr } from 'ember-decorators/data';
  *
  * export default Model.extend({
+ *   @attr({ defaultValue: () => [] }) personalItems;
+ * });
+ * ```
+ *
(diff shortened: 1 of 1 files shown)


akatov:
in principle, it's also possible to invoke `DS.attr` with `undefined` as the first argument (e.g. `myAttribute: DS.attr(undefined, { defaultValue: 5 })`) and it behaves in exactly the same way it would as if the first argument (type) was omitted altogether.
I managed to find a type for this with typescript 2.8's conditional types, however, I'm doubtful about including it in this (or a future) PR, because I think even though it works, calling it this way is ugly in the first place. What do you think?

buschtoens:
@akatov I don't have push access to this repo, sorry. I'm just an ember-decorators enthusiast. 😜

As to the `undefined` thing: You're right. While it's technically possible, it looks weird and the IntelliSense for it would probably be more confusing than helpful.

pzuraq:
Thank you!

Issue #232, opened 2018-05-10
Add @observes, @unobserves, @on, and @off decorators


pzuraq:
Add the following decorators:

* `@observes`
* `@unobserves`
* `@on`
* `@off`

These will bring the ember decorators project to feature parity with the Ember object-model. `@off` should remove any existing observers or event listeners.

allthesignals:
Would love to assist with this - is anyone else working on it?

pzuraq:
I've looked at it a little bit, `@observes` and `@on` should be easy enough, we can use `Ember.addObserver` and `Ember.addListener` directly. `@off` is trickier, we need to expose some functions in Ember to be able to get the listeners and observers that exist on a function:

* `getObservers`: https://github.com/emberjs/ember.js/blob/2e6edc5d27168bdc3873c5178e1ae3da52ac3cea/packages/ember-utils/lib/super.ts#L47
* `getListeners`: https://github.com/emberjs/ember.js/blob/2e6edc5d27168bdc3873c5178e1ae3da52ac3cea/packages/ember-utils/lib/super.ts#L59

machty:
@pzuraq I'm confused about what `@off` does? Is there some equivalent in the classic EC API? Is it `cancelOn` ?

pzuraq:
It solves the case where an observer or listener is overridden in a subclass:

```js
let Foo = EmberObject.extend({
  foo: observer()
});

let Bar = Foo.extend({
  foo() {

  }
});
```

In this case, the `foo` method should no longer be an observer on Bar. EmberObject accomplishes this when it when Mixins are being collapsed during the `proto` stage. 

Updating this to work with native classes would require some significant work, and given that it is not an incredibly commonly used feature (and that observers and listeners are increasingly recommended against) the `@off` decorator seemed like a better solution. Since observers and event listeners are technically both event listeners, it made sense to use just one decorator for both use cases.

lsg-richard:
Am quite interested in getting `@observes` working as its absence is preventing completing the conversion to ES6 classes in the app I manage. I'd be happy to take a stab at this if someone on the project could provide some initial direction. Also I'm on the community slack as `richard-nz`.

ed-asriyan:
@pzuraq I'm interested in getting `@observes` and `@off` in newer versions of ember-decorators. I've already implemented `@observes` in #267 and I would like to complete the feature by implementing `@off`. I understood its main idea, but I would like to clarify its usage.
 
```javascript
class Foo {
[...]

PR #267, opened 2018-08-05
[feat] Add @observes and @unobserves

ed-asriyan:
Implements `@observes` decorator, which was asked in #232.


pzuraq:
This is looking good so far! In order to merge this, we need to also make the opposite `@off` decorator so people can remove an observer in subclasses. It's an edge case, but we want to make sure folks can handle that case as well.

ed-asriyan:
@pzuraq rerun tests, please. Some of them failed on timeout.

pzuraq:
Thank you so much for adding this! Really appreciate it 😄

PR #270, opened 2018-08-20
[feat] Add @on and @off decorators

ed-asriyan:
Implements `@on` and `@off` decorators as described in #232.


pzuraq:
Thanks again for getting this and observers over the line! Will be releasing soon, I want to get a changelog incorporated really quick, then it'll be good to go 👍

PR #273, opened 2018-08-31
[FEAT] Adds throwOnComputedOverride build flag

pzuraq:
Adds `throwOnComputedOverride` build flag. I've tested locally to
confirm it works, and added a test for the assertion in the first place.
Will change the test once we change the default to false.


pzuraq:
In non-debug it should default to the setter behavior that computed has by default, we only add a setter if one doesn’t exist and in debug

rwjblue:
Gotcha. Thank you for walking through it...

Issue #64, opened 2016-03-20
Filter array based on another property?


nwhittaker:
Is there an equivalent way to filter an array when another property changes? Something that achieves the same thing as:

``` javascript
someItems: computed.filter('allItems', function(item) {
   return item.indexOf(this.get('filterText')) > -1;
}).property('filterText');
```

**EDIT:** Fix property to call `computed.filter()`, instead of `computed()`.

chrishowes:
+1

pzuraq:
Do you mean something more like one of these?

```
someItems: computed('allItems', 'filterText', function(items, filterText) {
  return items.filter(i => i.indexOf(filterText) > -1);
})
```

or

```
someItems: filter('allItems', function(item) {
   return item.indexOf(this.get('filterText')) > -1;
}).property('filterText');
```

Not sure how the original syntax would work

nwhittaker:
Sorry, yes. I meant `computed.filter()`. I'll edit the description.

pzuraq:
Will have to look more into how that worked originally, but we may be able to get that to work once we get composeable computeds working too

pzuraq:
We now have an open RFC which would address this issue by adding the ability to pass additional dependent keys to the affected computed macros, you can check it out [here](https://github.com/emberjs/rfcs/pull/375). Closing this issue for now.

PR #275, opened 2018-09-13
[feat] Adds computed modifiers

### LINKED ISSUE #272: Add @computed.readOnly and @computed.volatile
Currently, computed properties made with decorators are `readOnly` by default, and there is no way to have them be marked as `volatile` because undecorated native getters/setters are preferred. In order to create a general purpose codemod, we need to make sure that we have the exact same semantics as current computeds do. To do this, we need to make computeds clobberable by default, and add methods for invoking readOnly/volatile behavior.

pzuraq:
Adds the following modifiers for computed properties:

```js
@computed.readOnly()
@computed.volatile()
@(computed().property())
@(computed().meta())
```

These modifiers map to the built in modifiers in computed properties,
and are primarily being added to facilitate codemodding existing code.
Documentation has only been added for `.readOnly`, as the other
modifiers are _not_ recommended for use outside of updating existing
code.

This PR also removes the deprecation warning provided by the setter in
debug mode, unfortunately we cannot provide a setter _and_ allow users
to call `readOnly`. We'll have to wait for the official deprecation to
inform users that `readOnly` has been deprecated, if it comes along.

Fixes #272, #64


tomdale:
Can you explain the rationale for this API? Versus something like:

```ts
class MyClass {
  @computed @readOnly readOnlyProp;
  @computed @volatile volatileProp;
  @computed() @property() propertyProp;
  @computed() @meta() metaProp;
}
```

pzuraq:
We actually originally did have a `@readOnly` decorator like you described, but ran into a few issues with it that led to us changing to the chaining style here:

1. There is a very unfortunate naming conflict between the `readOnly` modifier and the `@readOnly` macro. Having two decorators with the same name but different behaviors depending on import path is really confusing, so that's a huge part of the reasoning. 

   Honestly, the `@readOnly` macro is also unfortunately named, once we deprecate `.readOnly()` it would make much more sense for it to be named `@reads` and for `@reads` to be named `@oneWay` again.. but that would probably also be confusing given the current state of macros.

2. Combinatorial decorator systems can get confusing for users. @mike-north can speak to this point, but it means more imports that they have to add, more individual decorators they have to learn, and oftentimes it means they have to remember a particular order to apply decorators in. 

3. It also is a major concern that decorator systems will collide - imagine a case where we have two different combinatorial decorator systems annotating the same function, figuring out which decorators belong to which system at a glance could be pretty hard.

All that said, I think in the future:
   
* We could make computed decorators order agnostic. This is pretty hard currently (but possible), and would be much easier in the future w/ stage 2/3 decorators.
* We will find out in time whether or not multiple decorator systems will actually exist, and whether or not chaining is a more or less confusing pattern compared to combinatorial decorators.
* Tooling will get better, to the point where adding an extra import and remembering paths won't be problematic to users.

This is also why I was hesitant to add a new `overridable()` modifier in the `readOnly` deprecation RFC - since we're deprecating `readOnly` and `volatile`, this API would truly be temporary, and we could wait to gather more information before adding back any modifiers and locking us into a particular pattern.

pzuraq:
This has been superceded

PR #283, opened 2018-09-27
[REFACTOR] Rename readOnly -> reads, reads/oneWay -> overridableReads

pzuraq:
The unfortunate naming conflict between the two different kinds of
`@readOnly` decorators is very confusing and makes it difficult to tell
what is going on without context. Now that overridability has been
deprecated, the naming makes even less sense.

This PR renames the `@readOnly` decorator to `@reads`, since it will
represent the default behavior in the new world (non-overridable one
directional alias). It also renames `@oneWay` to be `@overridableReads`
for clarity. This macro will likely continue to exist even in the
post-`readOnly` world, and continue to have the same functionality it
has now, so it makes sense to name it appropriately.

### DIFF
--- a/packages/-ember-decorators/tests/unit/object/macro-test.js
+++ b/packages/-ember-decorators/tests/unit/object/macro-test.js
@@ -26,10 +26,9 @@ import {
   none,
   not,
   notEmpty,
-  oneWay,
   or,
+  overridableReads,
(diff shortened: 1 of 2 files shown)


pzuraq:
cc @rwjblue and @mike-north

PR #290, opened 2018-10-18
fix(utils): only setup babel once

buschtoens:
I don't know, if the actual setup matters, because I did not create a reproduction repo yet, but it is quite convoluted. Basically it goes like this, all in one big monorepo:

```
host-app
+- lazy-loaded-engine
   +- some-addon
      +- @ember-decorators/utils
```
<details>
  <summary>Outdated, incorrect mumbling</summary>

~Somehow, if _and only if_ `some-addon` specifies non-`js` file extensions for `ember-cli-babel`, like shown below, it causes `@ember-decorators/utils` to be `included` multiple times into (presumably) `some-addon`.~

```js
// some-addon/index.js

module.exports = {
  name: '@clark/ember-native',

  options: {
    // as soon as this config key is present, things go wrong
    'ember-cli-babel': {
      extensions: ['js', 'ts'] // 'js' only causes no error
    }
  }
};
```
</details>

See https://github.com/ember-decorators/ember-decorators/pull/290#issuecomment-431169146 for updated information.

The following error is then thrown:

```
Build Error (broccoli-persistent-filter:Babel > [Babel: @clark/ember-native]) in @clark/ember-native/bridges/android.ts

Duplicate plugin/preset detected.
If you'd like to use two separate instances of a plugin,
they need separate names, e.g.

  plugins: [
    ['some-plugin', {}],
    ['some-plugin', {}, 'some unique name'],
  ]
    at assertNoDuplicates (/Users/janbuschtoens/clark/application/node_modules/@babel/core/lib/config/config-descriptors.js:206:13)
    at createDescriptors (/Users/janbuschtoens/clark/application/node_modules/@babel/core/lib/config/config-descriptors.js:114:3)
    at createPluginDescriptors (/Users/janbuschtoens/clark/application/node_modules/@babel/core/lib/config/config-descriptors.js:105:10)
    at alias (/Users/janbuschtoens/clark/application/node_modules/@babel/core/lib/config/config-descriptors.js:63:49)
    at cachedFunction (/Users/janbuschtoens/clark/application/node_modules/@babel/core/lib/config/caching.js:33:19)
    at plugins.plugins (/Users/janbuschtoens/clark/application/node_modules/@babel/core/lib/config/config-descriptors.js:28:77)
    at mergeChainOpts (/Users/janbuschtoens/clark/application/node_modules/@babel/core/lib/config/config-chain.js:314:26)
    at /Users/janbuschtoens/clark/application/node_modules/@babel/core/lib/config/config-chain.js:278:7
    at buildRootChain (/Users/janbuschtoens/clark/application/node_modules/@babel/core/lib/config/config-chain.js:68:29)
    at loadPrivatePartialConfig (/Users/janbuschtoens/clark/application/node_modules/@babel/core/lib/config/partial.js:85:55)
```

Some investigation has shown that this is caused by the following line being executed more than once for the including addon.

https://github.com/ember-decorators/ember-decorators/blob/d804943a8bd9bf079b532a12c2c4ae51aca6235c/packages/utils/index.js#L43

This PR ensures that `setupBabelPlugins(addon, options)` is only ever run once for any given `addon`.

Seeing that this only happens, if non-`js` file extensions are provided, this might not actually be the correct place to fix, but it works. I'll also raise an issue over at `ember-cli-babel`.

/cc @dfreeman @pzuraq @rwjblue

### DIFF
--- a/packages/utils/index.js
+++ b/packages/utils/index.js
@@ -1,4 +1,9 @@
+const didSetup = new Set;
+
 function setupBabelPlugins(addon, options) {
+  if (didSetup.has(addon)) return;
+  didSetup.add(addon);
+
   addon.options = addon.options || {};
   addon.options.babel = addon.options.babel || {};
 
@@ -40,7 +45,11 @@ function setupBabelPlugins(addon, options) {
     },
   };
 
-  plugins.push([require.resolve('babel-plugin-debug-macros'), pluginOptions]);
+  plugins.push([
+    require.resolve('babel-plugin-debug-macros'),
+    pluginOptions,
+    'ember-decorators-debug-macros'
+  ]);
 }
 
 module.exports = {


packages/utils/index.js
```diff
@@ -1,4 +1,9 @@
+const didSetup = new Set;
+
 function setupBabelPlugins(addon, options) {
+  if (didSetup.has(addon)) return;
```
pzuraq: This seems reasonable to me, I feel like we normally store this state on the addon (`addon._didSetupBabel`) but given that we're running this on parent's as well as the child I think it makes sense to use a `Set` here.

buschtoens:
Awesome, thanks for the quick merge!

packages/utils/index.js
buschtoens: FWIW, if somebody else stumbles across this, another potential problem with the `addon._didSetupBabel` pattern is that it only works for _one_ addon, that configures the host `includer`. If the host depends on two addons that both setup babel and happen to use the same marker name, one will not run.

Besides using a `Set`, in some cases it might be more appropriate to actually check whether or not certain plugins are included using [`ember-cli-babel-plugin-helpers`](https://github.com/dfreeman/ember-cli-babel-plugin-helpers).

buschtoens:
Seems like my solution was to naïve. I have now added `some-addon` as a dependency to `host-app` as well (in addition to `lazy-loaded-engine`):

```
host-app
+- lazy-loaded-engine
|  +- some-addon
|     +- @ember-decorators/utils
+- some-addon
   +- @ember-decorators/utils
```

And now I get the same errors again. :joy:

I'll submit a PR that actually checks that the specific plugin is not already in the plugins list.

PR #292, opened 2018-10-19
fix(utils): only install debug-macros plugin once

buschtoens:
https://github.com/ember-decorators/ember-decorators/pull/290#issuecomment-431366393

> Seems like my solution was too naïve. I have now added `some-addon` as a dependency to `host-app` as well (in addition to `lazy-loaded-engine`):
> 
> ```
> host-app
> +- lazy-loaded-engine
> |  +- some-addon
> |     +- @ember-decorators/utils
> +- some-addon
>    +- @ember-decorators/utils
> ```
> And now I get the same errors again. 😂
> 
> I'll submit a PR that actually checks that the specific plugin is not already in the plugins list.

Sorry for the back and forth. Should have probably implemented it like this in the first place anyway, but my brain was fried last night. 😵

### DIFF
--- a/packages/utils/index.js
+++ b/packages/utils/index.js
@@ -1,13 +1,13 @@
-const didSetup = new Set;
+const PLUGIN_NAME = 'ember-decorators-debug-macros';
 
 function setupBabelPlugins(addon, options) {
-  if (didSetup.has(addon)) return;
-  didSetup.add(addon);
-
   addon.options = addon.options || {};
   addon.options.babel = addon.options.babel || {};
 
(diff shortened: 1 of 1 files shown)


buschtoens:
Are we doing something weird with the `includer` thing or is ember-cli-babel a bit wonky? It feels a bit weird to me, that we have to employ these kinds of security measures.

pzuraq:
`ember-cli` has always been a bit wonky to me 😅 Sometimes addons get included more than once through different dependency trees, and when that happens I suspect we're running into this. Anyways, this looks better to me.

PR #298, opened 2018-11-06
Add PropertyDecorator type for `volatile` and `readOnly`. Fixes #297

### LINKED ISSUE #297: @readOnly should also be a PropertyDecorator
Right now (v3), the `@ember-decorators/object/readOnly` decorator is typed as a `MethodDecorator` because of its usage with `@computed()` getters. But `@readOnly` can also be used on properties, like so:

```ts
@readOnly
@or('_foo', '_bar')
readonly foo!: string;
```

Typescript should not show an error for this usage, which is supported and actually works.

HodofHod:
Fixes #297 

This usage was already supported, but wasn't typed. Types are now fixed and example usages have been added to the JSDocs. Also, the existing JSDoc usages were corrected.

### DIFF
--- a/packages/object/index.d.ts
+++ b/packages/object/index.d.ts
@@ -159,37 +159,53 @@ export function off(...eventNames: string[]): PropertyDecorator;
  *
  * ```javascript
  * import Component from '@ember/component';
- * import { computed, readOnly } from 'ember-decorators/object';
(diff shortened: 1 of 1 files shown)


pzuraq:
Thanks for the PR!

PR #300, opened 2018-11-07
add support for module-unified adapters

championswimmer:


championswimmer:
i think, better is to use this  https://github.com/ember-decorators/ember-decorators/pull/301 

it does not need index.js copy, simply uses the logic from ember-data

PR #302, opened 2018-11-07
fix package name in model blueprint

championswimmer:


pzuraq:
thanks for the PRs!

championswimmer:
@pzuraq my pleasure :)

When will we have patch version with these changes  ? (not to rush you, just asking)

pzuraq:
I'm currently working on some fixes upstream in `ember-cli-babel` and then I'm going to start working on upgrading `ember-decorators` to stage 2. These fixes will be out before that work is complete, since we want current users of v3 to be supported for a while, so probably in the next week or so

PR #305, opened 2018-11-12
docs(computed): fix sortDefinition documentation

makepanic:
Previously the jsdoc defined sortDefinition as possible string array. 
This is incorrect as tests check for usage as a string and [ember docs](https://emberjs.com/api/ember/3.5/functions/@ember%2Fobject%2Fcomputed/sort) describe the string sortDefinition to be a simple string.


pzuraq:
Thanks for the fix, really appreciate it!

PR #311, opened 2018-11-22
Get addon name from addon.name() or addon.name

### LINKED ISSUE #310: Updating to v3.1.0 gets an addon.name is not a function
I've just updated the library on my project and when I try to build the project I get this error:

``` 
addon.name is not a function
```
After some debugging I've found that it's this [line](https://github.com/ember-decorators/ember-decorators/blob/master/packages/utils/index.js#L58) that breaks the compilation and it doesn't let me to get the trace I needed (it was asking me to upgrade the library `ember-concurrency-decorators`).

Is it possible that some old addons had a different interface to getting their names? Anyway, I'm going to solve the problem and send a PR :smile:.

Gorzas:
Close #310

### DIFF
--- a/packages/utils/index.js
+++ b/packages/utils/index.js
@@ -21,6 +21,8 @@ function checkNeedsStage1Decorators(project) {
 function checkAddonsForStage1(project, addon) {
   if (!addon.addons || !addon.addons.length > 0) return false;
 
+  const addonName =  'function' === typeof addon.name ? addon.name() : addon.name;
+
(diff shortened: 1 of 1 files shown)


pzuraq:
This may be because of older versions of ember-cli. Either way, thanks for the fix!

josemarluedke:
I have been seeing this with Ember CLI 3.5.1. Hopefully we can get a release out soon with this fix.

pzuraq:
The fix for this has been released in v3.1.1

PR #318, opened 2018-11-27
fix(utils/decorator): transfer initializer, detect synthetic descriptors

buschtoens:
If we are dealing with legacy decorators and a decorated member has an `initializer`, the `initializer` is kept on the property descriptor and not transferred to the synthetic member descriptor. This PR deletes the `initializer` from the original descriptor and transfers it.

dd9348a0dca4e43af157d645f202bc92db544ce8 then transfers the initializer back to the property descriptor, just before we give it back to Babel.

Relevant part of the spec: [Field Decorator > Parameters](https://github.com/tc39/proposal-decorators/blob/master/METAPROGRAMMING.md#parameters)

This PR also adds a `toString` method to the synthetic member descriptor, so that `isStage2Descriptor` detects synthetic member descriptors as such to enable the pattern shown in https://github.com/ember-decorators/ember-decorators/issues/315#issuecomment-441720711 for legacy decorators as well.

### DIFF
--- a/packages/utils/addon/decorator.js
+++ b/packages/utils/addon/decorator.js
@@ -103,6 +103,13 @@ export function decorator(fn) {
           desc.finisher(target.prototype ? target : target.constructor);
         }
 
+        if (desc.initializer) {
+          // We've removed the initializer in `convertStage1ToStage2` to mirror
+          // the specs for field descriptor, but Babel 6 needs the initializer
+          // back on the property descriptor.
+          desc.descriptor.initializer = desc.initializer;
+        }
+
         return desc.descriptor;
       }
     }
--- a/packages/utils/addon/decorator.js
+++ b/packages/utils/addon/decorator.js
@@ -61,11 +61,16 @@ function convertStage1ToStage2(desc) {
     let kind = kindForDesc(desc);
     let placement = placementForKind(kind);
 
+    let { initializer } = descriptor;
+    delete descriptor.initializer;
+
     return {
       descriptor,
       key,
       kind,
       placement,
+      initializer,
+      toString: () => '[object Descriptor]',
     };
   } else {
     // Class decorator
--- a/packages/utils/addon/decorator.js
+++ b/packages/utils/addon/decorator.js
@@ -62,7 +62,6 @@ function convertStage1ToStage2(desc) {
     let placement = placementForKind(kind);
 
     let { initializer } = descriptor;
-    delete descriptor.initializer;
 
     return {
       descriptor,
@@ -93,7 +92,7 @@ export function decorator(fn) {
 
         fn(desc);
 
-        if (desc.finisher) {
+        if (typeof desc.finisher === 'function') {
           // Finishers are supposed to run at the end of class finalization,
           // but we don't get that with stage 1 transforms. We have to be careful
           // to make sure that we aren't doing any operations which would change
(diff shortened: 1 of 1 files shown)


packages/utils/addon/decorator.js
```diff
@@ -61,11 +61,16 @@ function convertStage1ToStage2(desc) {
...
 
+    let { initializer } = descriptor;
+    delete descriptor.initializer;
+
     return {
       descriptor,
       key,
       kind,
       placement,
+      initializer,
```
buschtoens: Should this field be conditionally set, so that `'initializer' in memberDescriptor === false`, if the original `descriptor` has no `initializer` in the first place?

packages/utils/addon/decorator.js
```diff
@@ -61,11 +61,16 @@ function convertStage1ToStage2(desc) {
     let kind = kindForDesc(desc);
     let placement = placementForKind(kind);
 
+    let { initializer } = descriptor;
+    delete descriptor.initializer;
```
pzuraq: I don't think we _need_ this `delete`, and my understanding is that generally it's best to avoid them when possible. Messes with the shape of objects and can cause deopts. I think it would be fine to pull the initializer off and set it back later.

packages/utils/addon/decorator.js
```diff
@@ -98,6 +103,13 @@ export function decorator(fn) {
           desc.finisher(target.prototype ? target : target.constructor);
         }
 
+        if (desc.initializer) {
```
pzuraq: Maybe this should be a `typeof` check to make sure it's a function

packages/utils/addon/decorator.js
pzuraq: I think we could get away with `typeof` checks, better to be precise anyways.

buschtoens: My primary concern was that Babel might not check for type, but mere presence. I recalled that Babel disallows the `initializer` property for some descriptors.

- https://github.com/babel/babel/blob/2bb24f996f025dccd26e1e5269d932efdeebc72d/packages/babel-helpers/src/helpers.js#L1629
- https://github.com/babel/babel/blob/2bb24f996f025dccd26e1e5269d932efdeebc72d/packages/babel-helpers/src/helpers.js#L1698

But looking at the implementation of [`_disallowProperty`](https://github.com/babel/babel/blob/2bb24f996f025dccd26e1e5269d932efdeebc72d/packages/babel-helpers/src/helpers.js#L1707-L1711) we're in the clear 😌

```js
function _disallowProperty(obj, name, objectType) {
  if (obj[name] !== undefined) {
    throw new TypeError(objectType + " can't have a ." + name + " property.");
  }
}
```

packages/utils/addon/decorator.js
buschtoens: Correct. We don't need it. I've removed it, so that users don't accidentally use or try to override this initializer in the wrong place.

I'll remove the `delete`.

packages/utils/addon/decorator.js
buschtoens: Should I update `desc.finished` for good measure as well?

https://github.com/ember-decorators/ember-decorators/blob/5489b24d7dfab3bbe71515ea719601672a000294/packages/utils/addon/decorator.js#L91-L99

pzuraq: Yup, that would be great!

PR #319, opened 2018-11-27
[data] Support types for ES6 Model subclass w/ @attr

mike-north:
#### Test Case

```ts
import { attr } from '@ember-decorators/data';

/// 🚨 This case will fail with current types 🚨///
class MyModel1 extends Model {
  @attr firstName: null 
}

class MyModel2 extends Model {
  @attr('string') firstName: null
}

class MyModel3 extends Model {
  @attr('string', { allowNull: false }) firstName: null
}

class MyModel4 extends Model {
  @attr() firstName: null
}
```

Currently all forms of the types for `attr` are functions that return decorators. This PR adds the form that uses `@attr` directly

### DIFF
--- a/packages/data/index.d.ts
+++ b/packages/data/index.d.ts
@@ -16,7 +16,7 @@ import ModelRegistry from "ember-data/types/registries/model";
  *
  * @function
  */
-export function attr(): PropertyDecorator;
+export function attr(target: Object, propertyKey: string | symbol): void;
 /**
  * Decorator that turns the property into an Ember Data attribute
  *


igorT:
We should do the same for `belongsTo` and `hasMany`

rwjblue:
> ```ts
> /// 🚨 This case will fail with current types 🚨///
> class MyModel1 extends Model {
>   @attr firstName: null 
> }
> 
> ```
Is there a way to write a test for this? I would imagine adding a single `.ts` file + tsc config that errors on invalid types would work, no?

pzuraq:
I think we could convert the tests themselves to Typescript at this point, since, TS is now using babel transforms.

Also I believe the type should overload, `attr` is both a decorator and a function that returns a decorator, depending on input.

mike-north:
For the time being, I've added support for testing types in a follow-on PR https://github.com/ember-decorators/ember-decorators/pull/320

This follows the same conventions (and uses the same tooling) as https://github.com/DefinitelyTyped/DefinitelyTyped

mike-north:
@pzuraq 

> Also I believe the type should overload, attr is both a decorator and a function that returns a decorator, depending on input.

This declaration
https://github.com/ember-decorators/ember-decorators/blob/acac6bb3a8525323e201b6c7c26ba9cf349e39cb/packages/data/index.d.ts#L35
Already handles the zero-argument function case (i.e., `@attr()`) since the options object is optional

pzuraq:
Ah, the doc comment above the diff made me think it was the only declaration. Looks good then!

PR #320, opened 2018-11-27
Type declaration testing

mike-north:
This PR adds support for writing tests against the hand-written type declarations that this package publishes


packages/controller/published-types/index.d.ts
```diff
@@ -57,6 +57,4 @@ export function controller(
  * @function
  * @param {String} [controllerName] - The name of the controller to inject. If not provided, the property name will be used
  */
-export function controller<K extends keyof Registry>(
-  name: K
-): PropertyDecorator;
+export function controller(name: keyof Registry): PropertyDecorator;
```
mike-north: This generic was unnecessary since the type parameter `K` was only used once

packages/data/published-types/index.d.ts
```diff
@@ -32,7 +34,8 @@ export function attr(): PropertyDecorator;
  * @function
  * @param {Object} [options] - Optional attribute options
  */
-export function attr(options?: DS.AttrOptions<any>): PropertyDecorator;
+// tslint:disable-next-line:unified-signatures
+export function attr(options?: DS.AttrOptions): PropertyDecorator;
```
mike-north: `DS.AttrOptions` already has a default generic type argument of `any`. There's no need to explicitly specify one

packages/service/published-types/index.d.ts
```diff
@@ -53,4 +55,4 @@ export function service(target: any, key: any, descriptor: PropertyDescriptor):
  * @function
  * @param {String} [serviceName] - The name of the service to inject. If not provided, the property name will be used
  */
-export function service<K extends keyof Registry>(serviceName: K): PropertyDecorator;
+export function service(serviceName: keyof Registry): PropertyDecorator;
```
mike-north: Type parameter that's only used once can be eliminated

mike-north:
This change allows us to write positive and negative test cases against the hand-written ambient types.

Any other means of detecting validating negative test cases is very hard, since some of them will end up causing the compilation to fail entirely (and you'll get no feedback).

It's now up to us to create `.ts` files in the `published-types` folder which represent all of the different use cases we wish to support

PR #326, opened 2018-12-02
[BUGFIX] Fix the * version check for ec-typescript (#325)

### LINKED ISSUE #325: Using ember-cli-typescript in an in-repo addon / engine causes compilation to fail.
Fails with the error:
```
Cannot read property '0' of null
```

The ember-cli-typescript version check doesn't allow for the "*" version typically used in engines.

https://github.com/ember-decorators/ember-decorators/blob/master/packages/utils/index.js#L75

kylenathan:
Fixes #325


kylenathan:
Closed in favour of #330

PR #328, opened 2018-12-03
Fix dependency checking

ef4:
Using `addon.root` here is perilous because it doesn't always point to the true NPM package root. Addons that customize `main` or `ember-addon.main` in package.json will have `addon.root` pointing at a subdirectory, and then you won't find package.json.

See https://github.com/ef4/ember-auto-import/issues/162 for an example.

This PR fixes the problem and also avoids re-reading the package.json, which is unnecessary.


pzuraq:
Thanks Ed, really appreciate it!

PR #331, opened 2018-12-04
Added code samples using init vs constructor

Caltor:
Added a code sample for using `init` using native classes when extending EmberObject and `constructor` when extending non EmberObjects.


pzuraq:
Thanks for the PR!

PR #336, opened 2018-12-11
fix(utils/decorator): handle desc return value

buschtoens:
While writing tests for legacy decorators for https://github.com/rwjblue/sparkles-component/pull/22, I noticed that the return value is ignored and that the descriptor needs to be mutated by reference.

I don't think that this is intentional, since we do the opposite for stage 2 decorators using the same callback function.

### DIFF
--- a/packages/-ember-decorators/tests/unit/utils/decorator-helpers-test.js
+++ b/packages/-ember-decorators/tests/unit/utils/decorator-helpers-test.js
@@ -1,12 +1,55 @@
 import { DEBUG } from '@glimmer/env';
 
 import { module, test } from 'qunit';
-import { decoratorWithParams, decoratorWithRequiredParams } from '@ember-decorators/utils/decorator';
+import {
+  decorator,
+  decoratorWithParams,
+  decoratorWithRequiredParams,
+} from '@ember-decorators/utils/decorator';
 
 module('decorator helpers', function() {
+  module('decorator', function() {
+    test('it works with returning a newly created member descriptor', function(assert) {
+      let decorate = decorator(desc => {
+        return {
+          ...desc,
+          descriptor: {
+            configurable: desc.configurable,
+            enumerable: desc.enumerable,
+            get() {
+              return 1337;
+            },
+          },
+        };
+      });
+
+      class Foo {
+        @decorate foo() {}
+      }
+
+      let foo = new Foo();
+      assert.strictEqual(foo.foo, 1337);
+    });
+
+    test('it warns about deprecated descriptor mutation by reference', function(assert) {
+      let decorate = decorator(desc => {
+        delete desc.descriptor.value;
+        delete desc.descriptor.writable;
+        desc.descriptor.get = () => 1337;
+      });
+
+      assert.expectDeprecation(() => {
+        class Foo {
+          @decorate foo() {}
+        }
+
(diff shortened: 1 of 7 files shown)


stefanpenner:
This looks like it needs a regression test.

buschtoens:
The build error seems to be unrelated. Can someone restart the build, please?

buschtoens:
There's quite some auto-formatting noise, but since there is a `.prettierrc.js`, I just kept it, to get it over with. I can undo it though.

packages/object/addon/index.js
```diff
@@ -188,17 +194,21 @@ export const observes = decoratorWithRequiredParams((desc, params) => {
...
-  desc.finisher = target => {
-    let { prototype } = target;
-
-    for (let path of params) {
-      removeObserver(prototype, path, null, desc.key);
-    }
+export const unobserves = decoratorWithRequiredParams(
+  (desc, params) => ({
+    ...desc,
+    finisher(target) {
```
buschtoens: Do we prefer this idiom over

```js
  (desc, params) => ({
    ...desc,
    finisher(target) {
      // ...
    }
  })
```

or

```js
  (desc, params) => {
    return {
      ...desc,
      finisher(target) {
        // ...
      }
    }
  }
```

Or is the `...spread` operator potentially creating bloat in transpiled code? Then a mix of directly mutating the desc and returning it might be preferable.

pzuraq: I don't have a strong preference style-wise here, I think `return` style is a little more readable at first, but once you internalize `({})` style it's not that much different and much less verbose so that's nice.

I think the spread operator won't add much in terms of bloat once we get externalized helpers working, but we are creating a new object instead of changing the original one, unsure how that could affect things perf-wise but in hot paths I know it can matter.

buschtoens: I would expect that at least for evergreen browsers the `{ ...old, override: foo }` pattern will be quite performant, because it is used massively by Redux users.

But since there is no _actual need_ to create a new object, mutating and returning the existing one is _probably_ a bit faster and just as good, though not as idiomatic.

I also don't mind either way. Maybe someone else would like to weigh in?

buschtoens: I've updated the code to use the mutate and return pattern, because it introduces less noise to this PR and is potentially a bit faster and means no code bloat even without externalized helpers.

packages/-ember-decorators/tests/unit/utils/decorator-helpers-test.js
```diff
@@ -1,12 +1,58 @@
...
+        @decorate foo() {}
+      }
+
+      let foo = new Foo();
+      assert.strictEqual(foo.foo, 1337);
+    });
+
+    (NEEDS_STAGE_1_DECORATORS
+      ? test
+      : skip)('it warns about deprecated descriptor mutation by reference', function(assert) {
```
buschtoens: Which pattern do you prefer?

```js
(NEEDS_STAGE_1_DECORATORS ? test : skip)('it ...', function(assert) { ... });
```

```js
if (NEEDS_STAGE_1_DECORATORS) {
  test('it ...', function(assert) { ... });
}
```

pzuraq:
Great job on this one!

PR #343, opened 2018-12-19
fix(package): move published-types to pkg root before publishing

### LINKED ISSUE #342: Vscode suggests wrong import path
Using the vscode suggestion feature ( or auto imports ) results in imports from the wrong path. For example:
`import { action } from "@ember-decorators/object/published-types";`
should be:
`import { action } from "@ember-decorators/object";`

screen shot 2018-12-15 at 2 47 59 pm

buschtoens:
Fixes #342.

This PR moves the `published-types/index.d.ts` one level up in the `prepack` script and undoes this in the `postpack` script.

> - `prepack`: run BEFORE a tarball is packed (on `npm pack`, `npm publish`, and when installing git dependencies)
> - `postpack`: Run AFTER the tarball has been generated and moved to its final destination.

— [npm docs](https://docs.npmjs.com/misc/scripts#description)

It also add the `published-types` directory to `.npmignore`, since it now does not contain anything useful to the package consumer any more (AFAICT).

### DIFF
--- a/packages/component/.npmignore
+++ b/packages/component/.npmignore
@@ -18,3 +18,6 @@ testem.js
 .node_modules.ember-try/
 bower.json.ember-try
 package.json.ember-try
+
+# The `index.d.ts` is moved one level up by the `prepack` script
+/published-types
--- a/packages/component/package.json
+++ b/packages/component/package.json
@@ -20,7 +20,9 @@
     "build": "ember build",
     "start": "echo \"Run serve in the main ember-decorators package\" && false",
     "test": "echo \"Run tests in the main ember-decorators package\" && false",
-    "dtslint": "dtslint published-types"
+    "dtslint": "dtslint published-types",
+    "prepack": "mv ./published-types/index.d.ts ./",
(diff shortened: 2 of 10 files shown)


buschtoens:
I've published [`@buschtoens/component@3.1.6-beta.1`](https://registry.npmjs.org/@buschtoens/component/-/component-3.1.6-beta.1.tgz) as a test and it works. VS Code uses the correct paths again.

mike-north:
There's a "light touch" way to fix this problem. 

1) Move `https://github.com/ember-decorators/ember-decorators/blob/master/packages/object/computed.d.ts` into the `published-types` folder

2) In the `package.json` for `@ember-decorators/object`, update the "types" property

```diff
-  "types": "published-types",
+  "types": "published-types/index.d.ts",
```


### Before 
screen shot 2018-12-19 at 3 15 52 pm


### After

screen shot 2018-12-19 at 3 15 17 pm

buschtoens:
@mike-north Perfect, thanks! Proper fix incoming. 😄

PR #344, opened 2018-12-19
Renovate config

### LINKED ISSUE #339: Setup Renovatebot or another automatic dependency bot
Something like [Renovatebot](https://renovatebot.com) or dependabot, should help us catch breakage like the recent dtslint failures

mike-north:
Fixes #339 

@rwjblue I see you suggested dependabot, and I think this does everything dependabot typically does, plus a few extra customizations for extra stability (i.e., treat SemVer minor releases that pass tests as safe to auto-merge, treat minor releases in TypeScript and `@types` packages as unsafe to auto merge).

Do you still feel Dependabot should be the preferred choice?

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,0 +1,73 @@
+{
+  "extends": [
+    "config:js-lib",
+    ":pinOnlyDevDependencies",
+    ":automergePatch",
+    ":automergeMinor",
+    ":semanticCommits"
+  ],
+  "timezone": "America/Los_Angeles",
+  "schedule": ["after 3am", "before 9am"],
+  "commitMessageTopic": "{{depName}}",
+  "commitMessage": "{{{commitMessagePrefix}}} {{{commitMessageTopic}}} {{{commitMessageSuffix}}}",
+  "commitMessageBody": "{{{commitMessageAction}}} {{{commitMessageExtra}}}",
+  "packageRules": [
+    {
+      "packageNames": ["typescript", "tslint", "typedoc", "dtslint"],
+      "groupName": "typescript packages",
+      "groupSlug": "typescript",
+      "minor": {
+        "automerge": false
+      }
+    },
+    {
+      "packagePatterns": ["^@types/"],
+      "groupName": "ambient types",
+      "groupSlug": "ambient-types",
+      "minor": {
+        "automerge": false
+      }
+    },
+    {
+      "groupName": "Linting packages",
+      "groupSlug": "linting",
+      "packageNames": ["babel-eslint", "eslint", "eslint-plugin-ember", "eslint-plugin-node"]
+    },
+    {
+      "packageNames": [
+        "broccoli-asset-rev",
+        "ember-ajax",
+        "ember-cli-babel",
+        "ember-cli-dependency-checker",
+        "ember-cli-eslint",
+        "ember-cli-eslint",
+        "ember-cli-fastboot",
+        "ember-cli-htmlbars-inline-precompile",
+        "ember-cli-htmlbars",
+        "ember-cli-inject-live-reload",
+        "ember-cli-qunit",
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -0,0 +1,73 @@
+{
+  "extends": [
+    "config:js-lib",
+    ":pinOnlyDevDependencies",
```
mike-north: `:pinOnlyDevDependencies` will lock `devDependencies` versions. For details RE: why renovate suggests doing this, they have a writeup [here](https://renovatebot.com/docs/dependency-pinning/)

If we want to leave SemVer ranges alone and just update the lock file, you can accept this suggestion 👇
```suggestion
    ":preserveSemverRanges",
```

renovate.json
```diff
@@ -0,0 +1,73 @@
...
+    ":automergePatch",
+    ":automergeMinor",
+    ":semanticCommits"
+  ],
+  "timezone": "America/Los_Angeles",
+  "schedule": ["after 3am", "before 9am"],
```
mike-north: This is useful to avoid potentially rate limiting your github user during business hours (I have this problem w/ renovate set up on ~60 libs)

renovate.json
```diff
@@ -0,0 +1,73 @@
...
+  "commitMessage": "{{{commitMessagePrefix}}} {{{commitMessageTopic}}} {{{commitMessageSuffix}}}",
+  "commitMessageBody": "{{{commitMessageAction}}} {{{commitMessageExtra}}}",
+  "packageRules": [
+    {
+      "packageNames": ["typescript", "tslint", "typedoc", "dtslint"],
+      "groupName": "typescript packages",
+      "groupSlug": "typescript",
+      "minor": {
+        "automerge": false
+      }
```
mike-north: This configuration basically says
* Group upgrades to any of `["typescript", "tslint", "typedoc", "dtslint"]` together in a single PR, and only trust _patch_ releases that pass tests enough to automatically merge.

renovate.json
```diff
@@ -0,0 +1,73 @@
...
+    },
+    {
+      "packagePatterns": ["^@types/"],
+      "groupName": "ambient types",
+      "groupSlug": "ambient-types",
+      "minor": {
+        "automerge": false
+      }
+    },
```
mike-north: Ambient type packages are published in batch. In the case of `@types/ember`, as many as ~24 packages may update at the same time -- you'll want them all in a single PR, rather than hoping that they'll each be upgraded successfully in isolation

renovate.json
```diff
@@ -0,0 +1,73 @@
...
+        "ember-source-channel-url",
+        "ember-try",
+        "eslint-plugin-ember",
+        "eslint-plugin-node",
+        "loader.js",
+        "qunit-dom"
+      ],
+      "groupName": "ember infrastructure",
+      "groupSlug": "ember-infra"
+    }
```
mike-north: When fixes are threaded through the `ember-cli/*` packages, often several releases are cut at once, and it's more efficient to treat this group of new versions as a single PR. 

In my packages, I have this set to automerge as well, mirroring the manual process I'd go through (i.e., if a major version of `ember-qunit` is cut and it passed my package's tests, there's almost 0 chance I'd give it much scrutiny before merging)

renovate.json
pzuraq: I'm generally more comfortable with updating the lockfile, since we can't pin transitive deps anyways, and we have tests which run with dependencies locked and then with them unlocked which helps us narrow down failures due to dependencies shifting. 

This _would_ help us narrow down dependency vs devDependency bugs a bit, but would also add noise in other ways I think (e.g. if a devDependency's transitive dep updated, could still cause errors)

pzuraq:
@mike-north Do we need to do anything else to setup the bot?

PR #346, opened 2018-12-19
Adds CODEOWNERS

pzuraq:


rwjblue:
FYI - In case others are curious about this 
https://help.github.com/articles/about-codeowners/

Issue #378, opened 2019-01-15
On 5.x decorated services cause TS error


mfeckie:
When using the `service` decorator in 5.x like this

```ts
import { inject as service } from '@ember-decorators/service';

export default class Foo extends Component {
  @service foo
}
```

TS generates this error `[ts] Unable to resolve signature of property decorator when called as an expression. [1240]`

If, however, we do

```ts
import { inject as service } from '@ember-decorators/service';

export default class Foo extends Component {
  @service() foo
}
```

The compiler is happy.

Not sure if this is a documentation tweak needed or an actual bug

pzuraq:
This is definitely a bug, we had tests for the JS behavior but not for the types. Got a PR out for it.

pzuraq:
This should be fixed in `v5.0.1`

PR #379, opened 2019-01-15
[BUGFIX] Fixes types for inject

pzuraq:
And does some general cleanup, using the generic types rather than
type signatures


pzuraq:
Fixes #378

packages/controller/published-types/controller-test.ts
```diff
@@ -5,11 +5,13 @@ declare module '@ember/controller' {
   interface Registry {
     ['x-foo']: Controller;
     ['x-bar']: Controller;
+    ['bazController']: Controller;
```
mike-north:
```suggestion
    bazController: Controller;
```

PR #380, opened 2019-01-16
[FEAT] Adds wrapComputed

pzuraq:
Adds a function that allows users to wrap existing computed properties, filling in the gaps left by `macro`. This is a temporary method that will become a noop in apps that support decorators.

### DIFF
--- a/packages/object/published-types/object-test.ts
+++ b/packages/object/published-types/object-test.ts
@@ -1,6 +1,7 @@
-import EmberObject from '@ember/object';
-import { computed, observes, unobserves, on, off, action } from '@ember-decorators/object';
+import EmberObject, { computed as emberComputed } from '@ember/object';
+import { computed, observes, unobserves, on, off, action, wrapComputed } from '@ember-decorators/object';
 import { alias } from '@ember-decorators/object/computed';
+import { test } from 'qunit';
 
 export default class UserProfileComponent extends EmberObject {
   first = 'John';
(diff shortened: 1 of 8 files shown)


packages/object/published-types/object-test.ts
```diff
@@ -27,6 +28,8 @@ export default class UserProfileComponent extends EmberObject {
     const [first, last] = value.split(' ');
     this.setProperties({ first, last });
   }
+
+  @wrapComputed(emberComputed(() => {})) test: any;
```
mike-north: could you add a test that ensures non-any types pass through the `wrapComputed` function properly?

mike-north: oh also, please add some tests involving retrieving the value of the computed property generated by `@wrapComputed`. `ComputedProperty` is a really sensitive and tricky area of `@types/ember`, due to the need to simultaneously support the full matrix of:

* Ember.Object.extend
* class extends Ember.Object
* this.get('foo')
* this.foo

I'd like to ensure that changes to `@types/ember` that could break your consumers -- either in defining CPs, wrapping them or consuming them --  will be picked up in this lib's test suite.

mike-north: > I did add an additional test below in classic class syntax to make sure that wrapComputed does not interfere with getters/setters, I think that should be sufficient for now.

LGTM

PR #383, opened 2019-01-17
[BUGFIX] Fixes observers

pzuraq:
Fixes two major problems with observers:

1. They didn't expand properties
2. They would completely break if used with non EmberObject based
   classes.

This PR fixes both issues, but only for stage 2 transforms. It works by
adding another field to the class that is a dummy field and finishes
chains during initialization.

This is very very hacky, but it should be possible to do this much
better with `kind = 'initializer'` descriptors in stage 3, once they're
implemented. It shouldn't mess with shape or anything (I don't think),
will just be a bit ugly. The prop is non-enumerable too, so should be
very little impact.

### DIFF
--- a/packages/-ember-decorators/tests/unit/object/computed-test.js
+++ b/packages/-ember-decorators/tests/unit/object/computed-test.js
@@ -425,7 +425,7 @@ module('@computed', function() {
   module('modifiers', function() {
     test('volatile', function(assert) {
       assert.expect(2);
-      class Foo {
+      class Foo extends EmberObject {
         _count = 0;
 
         @(computed('first').volatile())
--- a/packages/object/addon/index.js
+++ b/packages/object/addon/index.js
@@ -1,7 +1,9 @@
+import { NEEDS_STAGE_1_DECORATORS } from 'ember-decorators-flags';
 import { assert } from '@ember/debug';
 
-import { computed as emberComputed } from '@ember/object';
-import ComputedProperty from '@ember/object/computed';
+import Ember from 'ember';
+import EmberObject, { computed as emberComputed } from '@ember/object';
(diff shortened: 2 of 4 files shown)


packages/-ember-decorators/tests/unit/object/computed-test.js
```diff
@@ -425,7 +425,7 @@ module('@computed', function() {
   module('modifiers', function() {
     test('volatile', function(assert) {
       assert.expect(2);
-      class Foo {
```
rwjblue: Why do we need observers in this test? It isn't obvious to me...

packages/object/addon/index.js
```diff
@@ -276,17 +279,50 @@ export const wrapComputed = computedDecoratorWithParams((desc, params) => {
...
+      placement: 'own',
+      key: '__EMBER_DECORATORS_FINISH_CHAINS__',
+      descriptor: {
+        enumerable: false,
+        writable: true,
+        configurable: true,
+      },
+      initializer() {
+        if (!CHAINS_FINISHED.has(this)) {
+          Ember.finishChains(Ember.meta(this));
```
rwjblue: Hmm, we _don't_ need to do this if working on `Ember.Object`'s though (AFAICT). Can we avoid the double call to `.finishChains` when working on Ember.Object's?

packages/-ember-decorators/tests/unit/object/computed-test.js
pzuraq: Ensuring that `volatile` works correctly. One of the things it does is make a property unobservable, so observers won't trigger.

packages/object/addon/index.js
pzuraq: Yes, definitely, I actually had that a little while ago, must not have committed for some reason...

PR #386, opened 2019-01-17
[BUGFIX] Ensure we only define one super special prop per class

pzuraq:
This (somewhat hackily) ensures we don't add the initializer prop more than once. The technique should be safe, we shouldn't have to deal with re-entry (how would you even define a class while defining a class?).

### DIFF
--- a/packages/object/addon/index.js
+++ b/packages/object/addon/index.js
@@ -278,6 +278,7 @@ export const wrapComputed = computedDecoratorWithParams((desc, params) => {
   @function
   @param {...String} propertyNames - Names of the properties that trigger the function
  */
+let hasChainsFinished = false;
 const CHAINS_FINISHED = new WeakMap();
 
 export const observes = decoratorWithRequiredParams((desc, params) => {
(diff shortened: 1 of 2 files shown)


packages/object/addon/index.js
```diff
@@ -286,28 +287,37 @@ export const observes = decoratorWithRequiredParams((desc, params) => {
...
-
-          CHAINS_FINISHED.set(this, true);
+  // hasChainsFinished will be set to true when the first initializer field
+  // is added during class evaluation, then back to false when the finishers
+  // run. This is not ideal, but shouldn't be an issue in the future when we
+  // have the ability to add arbitrary initializers that don't have to be fields
+  if (hasChainsFinished === false) {
+    hasChainsFinished = true;
+
+    // hackity hackity haaaaaack
```
NullVoxPopuli: :+1:

jacobq: :notes: To the tune of [*Yakety Yak*](https://en.wikipedia.org/wiki/Yakety_Yak) no doubt...

PR #388, opened 2019-01-23
[BUGFIX] Ensure app options are used for deco order

pzuraq:


packages/babel-transforms/index.js
```diff
@@ -8,7 +8,7 @@ module.exports = {
     this._super.included.apply(this, arguments);
 
     // Create parent options, if they do not exist
-    let parentOptions = (parent.options = parent.options || {});
+    let parentOptions = (this.app || this.parent).options || {}
```
rwjblue: 👍 - This looks good (as a mechanism to get the options)

PR #424, opened 2019-04-03
fix(deprecation): prefer _setClassicDecorator over _setComputedDecorator

josemarluedke:
This fixes the deprecation added in Ember by PR https://github.com/emberjs/ember.js/pull/17833.

cc/ @pzuraq @rwjblue


pzuraq:
Thanks for the fix @josemarluedke!

PR #425, opened 2019-04-03
[BUGFIX] Removes optional stage 1 code checks

pzuraq:
We used to actually check if stage 1 was necessary, but
since we are now reverting to stage one there is no point.
We'll remove the stage 2 code in v6 or v7.


rwjblue:
In general, seems good. Not sure whats up with CI.

pzuraq:
IE11 is what's up, I always forget that local targets are different than CI 😛 should be good now

PR #440, opened 2019-04-25
fix(utils): [Edge 18] ensure function prototype

### LINKED ISSUE #417: v5.2 MS Edge 18: Object doesn't support property or method 'apply'
We've encountered a failure case with ember-decorators in combination with ember-cli-typescript@1.5.0

The error does not happen in Edge 17 or any other browser that we've seen.

`Object doesn't support property or method 'apply'` is the error and cause the app not to boot.  We've been able to reproduce in two seperate codebases.

The error seems to occur at https://github.com/ember-decorators/ember-decorators/blob/master/packages/utils/addon/decorator.js#L89

The item returned by `decorate(fn)` is not correctly defined for some reason.

I've put together a repo with a reproducible failure case and detailed all the information I've found form poking about so far, in the README.

https://github.com/mfeckie/decorator-edge-repro

I don't have sufficient knowledge of the internals of decorators to be able to offer a fix, but happy to help with more information as necessary.

buschtoens:
Fixes #417.

Apparently (starting with Edge 18?) Edge's JS engine performs some broken optimizations for the return value of `decorator`. The returned function somehow gets stripped off its `Function` prototype and instead gets `Object`.

This fix ensures the `Function` prototype for Edge.

### DIFF
--- a/packages/utils/addon/decorator.js
+++ b/packages/utils/addon/decorator.js
@@ -6,8 +6,14 @@ import {
   convertStage1ToStage2,
 } from './-private/class-field-descriptor';
 
+const IS_EDGE = /Edge\/\d/.test(window.navigator.userAgent);
+
+const ensureFunctionPrototype = IS_EDGE
+  ? fn => Object.setPrototypeOf(fn, Function.prototype)
+  : fn => fn;
+
 export function decorator(fn) {
-  return function(...params) {
+  return ensureFunctionPrototype(function(...params) {
     if (isStage2FieldDescriptor(params)) {
       deprecate(
         'You are using the stage 2 decorator trasforms (@ember-decorators/babel-transforms v3-v5). Ember has officially adopted the stage 1 transforms instead. The stage 2 transforms will not be supported in ember-decorators v6. You can update `ember-cli-babel` to the latest version (at least 7.7.3) and remove @ember-decorators/babel-transforms from your app/addon.',
@@ -46,7 +52,7 @@ export function decorator(fn) {
 
       return desc.descriptor;
     }
-  };
+  });
 }
 
 /**


mfeckie:
Are there any blockers to getting this merged?  (awesome that you've found a workable fix)

buschtoens:
CI is acting up for unrelated reasons for the `ember-beta` and `ember-canary` build, it seems. I restarted them, but they still fail. `ember-cli-dependency-checker` for some reason cannot find the `ember-data` package. It might not be installed.

I'll try to fix this later today. Then somebody with npm publish access (@rwjblue and @pzuraq) can publish.

mfeckie:
How we going here peeps?

pzuraq:
@mfeckie sorry, it just hasn't been a priority since we have a lot of work to do to ship decorators/tracked properties upstream, and the recommendation is to upgrade to v6 which shouldn't have this issue 😕 If Jan can get tests passing and handle the deployment, I'm 👍 on landing this, but I'm not going to have time to look into it myself.

buschtoens:
Unfortunately we already migrated all of our apps to v6 as well. 🙊

mfeckie:
Yeah, I think that's what we're likely to do ... time permitting :)

packages/utils/addon/decorator.js
```diff
@@ -6,8 +6,14 @@ import {
   convertStage1ToStage2,
 } from './-private/class-field-descriptor';
 
+const IS_EDGE = /Edge\/\d/.test(window.navigator.userAgent);
```
thec0keman: should this be wrapped with a FastBoot check?

rwjblue: Yes

buschtoens:
```suggestion
const IS_EDGE = typeof FastBoot === 'undefined' && /Edge\/\d/.test(window.navigator.userAgent);
```

rwjblue:
Just wanted to clarify here, this PR targets the v5 branch but I'm not 100% sure why? Is it only applicable to v5?

buschtoens:
@rwjblue that is correct. This issue only occurs with v5.

wytlytningNZ:
Hey folks, is there any chance we could get some movement on this? 🙏 I would be more than happy to help out where possible if time is the main blocker.

pzuraq:
@wytlytningNZ unfortunately none of the maintainers have time to work on this issue, since it doesn't affect the current major release. Our recommendation is to upgrade to v6. 

If you would like to take over this PR and get tests passing, that would be really helpful. At the least it would mean you could pin your `package.json` to a commit, if we still don't have time to merge and publish a patch version.

buschtoens:
You can also use this hotfix: https://github.com/ember-decorators/ember-decorators/issues/417#issuecomment-486703338

But I can only reiterate on @pzuraq's point. Rather try upgrading to v6 as quickly as possible, to avoid being locked in to and old and unmaintained state of the Ember ecosystem.

Issue #327, opened 2018-12-03
isStage1Descriptor: false positive, if passed a function


buschtoens:
If you pass a `function() {}` as the only argument to a decorator, [`isStage1Descriptor`](https://github.com/ember-decorators/ember-decorators/blob/f6259d7e38e25f60d249e8771f340b89e50dc42f/packages/utils/addon/decorator.js#L15-L38) returns a false positive.

```ts
// handler is a function
const myDecorator = decoratorWithRequiredParams((desc, [handler]) => desc);

// breaks
@myDecorator(function() {})
class Foo {}

// works
@myDecorator(() => {})
class Foo {}
```

https://github.com/ember-decorators/ember-decorators/blob/f6259d7e38e25f60d249e8771f340b89e50dc42f/packages/utils/addon/decorator.js#L31-L35

Do you think that there is any way to reliably tell a `function() {}` apart from a `class {}`?

Funny coincidence, that ["How Does React Tell a Class from a Function?" by Dan Abramov](https://overreacted.io/how-does-react-tell-a-class-from-a-function/) is trending on HN now. 😄

buschtoens:
Taking the [babel class transforms](https://babeljs.io/en/repl#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=true&code_lz=MYewdgzgLgBAYjAvDYAbAhhCMDeAoGGAWwFMoALEAEwAoBKXAXz2b1EhFRIDpUQBzGnDpA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=es2015%2Ces2016%2Ces2017%2Cstage-0%2Cstage-1%2Cstage-2%2Cstage-3%2Ces2015-loose%2Ctypescript&prettier=false&targets=&version=7.0.0-beta.50&envVersion=7.0.0-beta.51) into account, I don't think there is any way to tell the difference between a `function() {}` and a `class {}`.

We could either do it on a best effort basis, which will still fail in certain cases, or just outright "ban" users from passing a single function as a parameter.

Some ideas for heuristics:

- We can check the length of the prototype chain. `function() {}` and `class {}` have the same length, but derived classes have longer chains: One extra link per `extends`.
- We can check whether `target.prototype instanceof Ember.Object`. This would reliably detect all classes that extend Ember framework classes.

[...]

PR #451, opened 2019-07-01
fix(component): `@layout` accept template factory

buschtoens:
https://github.com/emberjs/ember.js/pull/18096

This PR changed how templates get transpiled under the hood.
The export from templates are now factory functions which expect to be passed the `owner`.


rwjblue:
CI is failing due to using Node 6:

```
@ember-data/-build-infra/node_modules/broccoli-rollup/dist/index.js:31
    async build() {
          ^^^^^
```

buschtoens:
Should we pin the Ember Data dependency for now and do a major release later that drops Node 6?

packages/component/addon/index.js
```diff
@@ -201,7 +201,9 @@ export const layout = decoratorWithRequiredParams((target, params) => {
   );
   assert(
     `The @layout decorator must be provided a template, received: ${template}`,
-    (() => typeof template === 'object' && typeof template.indexOf === 'undefined')()
+    (() =>
+      typeof template === 'function' ||
+      (typeof template === 'object' && typeof template.indexOf === 'undefined'))()
```
buschtoens: btw is there any reason this is an IIFE? If not, I'd remove that in this PR as well.

buschtoens:
I am sure that the [`typescript@next` error](https://travis-ci.org/ember-decorators/ember-decorators/jobs/554125248#L550-L552) is caused by https://github.com/microsoft/TypeScript/pull/30790.

```
Error: Errors in typescript@next for external dependencies:
../../../node_modules/@types/node/index.d.ts(187,11): error TS2300: Duplicate identifier 'IteratorResult'.
../../../node_modules/dtslint/typescript-installs/next/node_modules/typescript/lib/lib.es2015.iterable.d.ts(41,6): error TS2300: Duplicate identifier 'IteratorResult'.
```

I reported it here: https://github.com/microsoft/TypeScript/pull/30790#issuecomment-508411128

buschtoens:
Oh come on FFS, how hard can it be to get a freaking one-line change merged? 😂

`ember-cli-addon-docs` and all the dependencies it brings in are just causing _sooo_ many issues. I'll open another PR to get `master` clean first, when I can find some time.

buschtoens:
Alright, I discovered another issue with https://github.com/emberjs/ember.js/pull/18096 in combination with `@layout` in production builds. It's a reprise of #327.

In development builds or builds that only target browsers that support arrow functions, the returned `TemplateFactory` and thus first argument passed to `@layout` is an arrow function, e.g. `owner => { ... }`.

Importantly, arrow functions have no `prototype` key, while regular functions (`function(owner) { ... }`) do.

https://github.com/ember-decorators/ember-decorators/blob/9f6af93075f1c2c216759e4e763ab487d382958b/packages/utils/addon/-private/class-field-descriptor.js#L1-L10

This causes the above code to report a false positive in production builds an lets the `@layout` decorator fail with:

> Error: Assertion Failed: The `@layout` decorator requires parameters

Since [`decoratorWithRequiredParams`](https://github.com/ember-decorators/ember-decorators/blob/9f6af93075f1c2c216759e4e763ab487d382958b/packages/utils/addon/decorator.js#L50-L59) thinks the decorator is invoked like:

```ts
@layout
class {}

// instead of
@layout(template)
class {}
```

I suggest we stop using `decoratorWithRequiredParams` for `@layout`. Unfortunately we use a check here then, but I don't see any other way to fix this.

rwjblue:
Ya, that seems fine to me...

josemarluedke:
@buschtoens @rwjblue  Is there anything that I can help with to move this forward? I'm currently blocked on upgrading Ember Canary because of the issue this solves.

packages/component/addon/index.js
rwjblue: No not particularly. I'd only use an IIFE in order to simplify the boolean logic, e.g.:

```js
assert(
  'whatever',
  (() => {
    if (condition) { return true; }
    if (otherThing) { return true; }
    return false;
  })()
```

It would probably make this specific case easier to think about, but isn't a giant deal.

rwjblue:
AFAIK the blocker is just implementing the change @buschtoens suggested

buschtoens:
I fixed the `decoratorWithRequiredParams` usage and removed the IIFE, but I expect CI to still fail spectacularly, because of all the other unrelated issues 🙈

rwjblue:
@pzuraq - Thoughts on the path forward here/

pzuraq:
I think the solution looks good, I'm not actually sure what the test failures on Canary are, it looks like they're coming from pretty deep inside the VM O_o 

Also unsure what's going on with production tests. I can try to take a look sometime next week, going to be busy for the remainder of this week. I also think we could make canary tests optional at this point, since we don't expect much to be changing with regards to the object model in the near future. Ideally we wouldn't, but we could to get unblocked.

buschtoens:
The canary failure is caused by https://github.com/emberjs/ember-test-helpers/pull/677

rwjblue:
Ya, could be but I thought I fixed those issues in ember so it didn't require changes in addons (e.g. didn't need to upgrade `@ember/test-helpers`).

josemarluedke:
Failures are agast "canary" and "production", could it be that when the tests ran, the build didn't pick up the changes that @rwjblue is referring as fixed? 

Error: `Cannot read property 'lookup' of undefined`

The error does seem to be related, maybe some can restart the tests to see if it passes now? 

Looking forward to having this fixed.

rwjblue:
The production build failure was not related to the lookup failure (it was `Error: Browser failed to connect within 30s. testem.js not loaded?`).

Apparently my fix in ember-source only fixed `setupRenderingTest` style tests, not `moduleForComponent`. I haven't had a chance to dig into that yet though. It is blocking a couple legit bug fixes over in @ember/test-helpers though.

rwjblue:
A plausible path forward here would be to run the codemod to migrate to setupRenderingTest to work around the bug in moduleForComponent.

rwjblue:
OK, now ready for a rebase...

rwjblue:
Rebased and merged in #454

PR #456, opened 2019-08-12
[FEATURE] Adds the @classNameBindings and @attributeBindings decorators

### LINKED ISSUE #450: Add @classNameBindings and @attributeBindings class decorators
Currently, it's not possible to directly convert certain bindings because dynamic properties cannot be decorated. Adding two class decorators to specify the values directly would prevent this from being an issue. This will also be a much easier to target output for the codemod.

pzuraq:
Adds decorators for classNameBindings and attributeBindings directly. This
makes it easier to have a 1-1 conversion from classic classes to native
classes with classic components.

Fixes #450


pzuraq:
Released in v6.1.0

PR #461, opened 2019-09-06
[BUGFIX] Adds types for new decorators

pzuraq:


packages/component/published-types/index.d.ts
```diff
@@ -191,6 +191,32 @@ export function className(trueValue?: string, falseValue?: string): PropertyDeco
...
 
+/**
+  Class decorator which specifies the class name bindings to be applied to a
+  component. This replaces the `classNameBindings` property on components in the
+  traditional Ember object model.
+  ```js
+  @classNameBindings('aDynamicProperty:truthy-class-name:falsy-class-name')
+  export default class ClassNamesDemoComponent extends Component {}
+  ```
+  @function
```
mike-north: do we use yuidoc to document this?

pzuraq:
To be fair, these docs are copied directly from the plain JS code, so if the use cases are different we could alter them to be different and provide a better DX for editors/TS users. At the least, `@function` could be removed, it's only there because ESDoc can't tell what the value is (it thinks it's a variable 😕)

The decision to use plain markdown for examples and documentation in ec-addon-docs was driven by real-world use cases:

1. Ability to interleave examples with documentation, to be able to provide more thorough API docs
2. Ability to specify language, so docs for APIs that require examples in multiple languages (in particular, component documentation) can be thoroughly documented

If `@example` in modern documenters can handle these use cases, totally for updating them. This also may not be as big of an issue if we start using Storybook or something like it for documenting components/other common multi-language primitives, and used doc comments _just_ for plain JS APIs

PR #462, opened 2019-09-20
Refresh yarn.lock

rwjblue:


rwjblue:
I was getting errors locally when using Node 12 (because we needed an fsevents update).

PR #496, opened 2026-04-27
Setup gh pages deploy

BobrImperator:
Depends on #494 

- Adds a CI workflow and a complementary prepare-deploy.mjs script for pushing Docs to gh-pages.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,52 @@
+name: Deploy Docs
+
+on:
+  push:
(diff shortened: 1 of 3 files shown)


mansona (approved):
I'm not too happy with the ember-cli-addon-docs setup in general but I'm going to merge this and test in production 🙈 I tried to verify this locally but the strange behaviour of github pages means that we can't even verify the deployment

Once we do another release we can verify that this updated and fix it then 👍

PR #503, opened 2026-06-08
Migrate @ember-decorators/component to a v2 addon

BobrImperator:
Related to https://github.com/ember-decorators/ember-decorators/pull/494

- Migrates @ember-decorators/utils to a v2 addon.


packages/component/blueprints/classic-component-addon/files/__root__/__path__/__name__.js
mansona, line 1: So... if we're deciding to remove the blueprints that's totally fine, but the docs would have to change. https://ember-decorators.github.io/ember-decorators/docs says that you call `ember install ember-decorators` and it "installs the packages", that won't be true if we remove all the blueprints (and we need to find all the docs that refer to generating new things and update it)

I personally think that the simplest thing here is to just leave the blueprints where they are 🤷 blueprints work fine in v2 addons without changing anything, so it will make the PR diff smaller anyway 😂 and if we want to remove blueprints we can do that with a follow-up PR later 👍

mansona (changes requested):

mansona (approved):
looks great 🎉

